PR #29, opened 2023-12-01
1.1.0-e-01Dec2023: add single container architecture and relevant changes to suppo…

scrthq:
# 1.1.0-e-01Dec2023

* Introduced single-container architecture via single Dockerfile in the repo root
    * Updated `utils/*.sh` and `ash` shell scripts to support running within a single container
    * Added new `ash_helpers.{sh,ps1}` scripts to support building and running the new container image
* Changed CDK Nag scanning to use TypeScript instead of Python in order to reduce the number of dependencies
* Changed identification of files to scan from `find` to `git ls-files` for Git repositories in order to reduce the number of files scanned and to avoid scanning files that are not tracked by Git
* Updated the multi-container Dockerfiles to be compatible with the script updates and retain backwards compatibility
* Updated ASH documentation and README content to reflect the changes and improve the user experience
* Added simple image build workflow configured as a required status check for PRs


utils/cfn-to-cdk/cfn_to_cdk/cfn_to_cdk_stack.py
```diff
@@ -8,7 +8,6 @@ class CfnToCdkStack(cdk.Stack):
     def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
         super().__init__(scope, construct_id, **kwargs)
 
-    
-        template0 = cfn_inc.CfnInclude(self, "/app/test.yaml",  
+
```
climbertjh2, line 11: With the switch to using the TypeScript cdk-nag-scan method, I think this file can be deleted rather than updated.

climbertjh2 (commented):
I left one comment  - optional to fix it in this (large) update.  The rest looks good to me - so I'm approving.

Issue #33, opened 2023-12-20
Issue with Dockerfile-cdk, and returning with '1' with latest version -> Dockerfile-cdk :   1


upi360:
I was trying to scan my code in cloud9 cdk environment and getting the below error message at end of the scan. I am using the latest version of the ASH  - 1.1.0-e-01Dec2023

=> ERROR [5/6] RUN cd /utils/cdk-nag-scan &&     npm install              0.3s
------
 > [5/6] RUN cd /utils/cdk-nag-scan &&     npm install:
#0 0.275 /bin/sh: 1: cd: can't cd to /utils/cdk-nag-scan
------
Dockerfile-cdk:18
--------------------
  17 |     
  18 | >>> RUN cd /utils/cdk-nag-scan && \
  19 | >>>     npm install
  20 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c cd /utils/cdk-nag-scan &&     npm install" did not complete successfully: exit code: 2
Unable to find image 'scan-11607:latest' locally
docker: Error response from daemon: pull access denied for scan-11607, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
Dockerfile Dockerfile-cdk returned 125
Error response from daemon: No such container: scan-11607

scrthq:
@upi360 Working on this now, thank you for raising!

scrthq:
hi @upi360 - This should be resolved now in `main` by PR #34 . Please let us know if you are still having issues by reopening this ticket if anything persists. Thank you, have a nice day!

PR #34, opened 2024-02-01
fix(#33): revert npm install on multi-container-arch to resolve cd issue

scrthq:
*Issue #, if available:* #33

*Description of changes:*

- Restored `npm install` on `utils/cdk-docker-execute.sh`, also swapped `--quiet` for `--silent`
- Removed `npm install` from `helper_dockerfiles/Dockerfile-cdk` as it runs before utils are mounted in multi-container architecture
- Added `npm install -g npm` to root `Dockerfile` to ensure `npm` is update to latest during single-container build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -121,8 +121,8 @@ COPY . /ash/
 # Install CDK Nag stub dependencies
 #
 # Update NPM to latest
-RUN npm install -g npm
-RUN cd /ash/utils/cdk-nag-scan && \
+RUN npm install -g npm && \
+    cd /ash/utils/cdk-nag-scan && \
     npm install --quiet
 
 #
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -120,8 +120,10 @@ COPY . /ash/
 #
 # Install CDK Nag stub dependencies
 #
+# Update NPM to latest
+RUN npm install -g npm
```
climbertjh2, line 124: Does this need to be a separate `RUN` step (resulting in an additional layer being created)?

Could the `npm install -g npm` be the first command run before the `cd ... && ...` commands in the next  `RUN` step - similar to the pattern for `yum update && yum upgrade -y` pattern in many `Dockerfile`s?

scrthq: @climbertjh2 we can add it on the same run call, I'll update that now. Not a bad suggestion

climbertjh2 (approved):
Looks good!

PR #35, opened 2024-02-01
Fix malapropism

john-aws:
Suspect this should read "overwritten", not "overridden".


scrthq (approved):
Nice catch, thank you for the PR!

PR #43, opened 2024-03-06
feat(arch): updated Dockerfile and ash_helpers.* to allow arm64 builds

scrthq:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,13 +3,8 @@
 #
 # Enable BASE_IMAGE as an overrideable ARG for proxy cache + private registry support
 #
-ARG BASE_IMAGE=public.ecr.aws/bitnami/python:3.10
-#
-# Platform is set to `linux/amd64` as that is the only platform
-# the current BASE_IMAGE supports while we identify a solution
-# for supporting arm64 as well.
-#
-FROM --platform=linux/amd64 ${BASE_IMAGE}
+ARG BASE_IMAGE=public.ecr.aws/docker/library/python:3.10-bullseye
+FROM ${BASE_IMAGE}
 
 #
 # Setting timezone in the container to UTC to ensure logged times are universal.
--- a/utils/ash_helpers.ps1
+++ b/utils/ash_helpers.ps1
@@ -124,7 +124,6 @@ function Invoke-ASH {
                 $buildCmd = @(
                     $RESOLVED_OCI_RUNNER
                     'build'
-                    '--platform linux/amd64'
                     '-t'
                     $AshImageName
                     "'$ashRoot'"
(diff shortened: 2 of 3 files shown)


climbertjh2 (approved):
Changes look good to me.

My only comment is that the Github action step which runs ash on itself seems to write the output folder "under" the source source folder.  That might confuse some of the scanners by having files in the "source" area changing mid-scan".

Approving.

scrthq:
> Changes look good to me.
> 
> My only comment is that the Github action step which runs ash on itself seems to write the output folder "under" the source source folder. That might confuse some of the scanners by having files in the "source" area changing mid-scan".
> 
> Approving.

Thanks, @climbertjh2 ! We can write it out somewhere else if it makes sense, but that's purely for this repo's pipeline. If we see it in the scan, I agree it should go somewhere else though. Easy to fix, will check it out and push one more change if it's causing trouble.

climbertjh2 (approved):
still looks OK to me.  Approving.

ash
```diff
@@ -1,624 +1,129 @@
...
+    echo "Resolved OCI_RUNNER to: ${RESOLVED_OCI_RUNNER}"
+
+    # Build the image if the --no-build flag is not set
+    if [ "${NO_BUILD}" = "NO" ]; then
+      ${RESOLVED_OCI_RUNNER} build \
+        --tag ${ASH_IMAGE_NAME} \
+        --file "${ASH_ROOT_DIR}/Dockerfile" \
+        ${DOCKER_EXTRA_ARGS} \
+        "${ASH_ROOT_DIR}"
+      echo $build_cmd
```
orsifacundo, line 108: This echo should be gone

ash
```diff
@@ -1,624 +1,129 @@
...
-TOTAL_EXECUTION=$((END_TIME-START_TIME))
-
-echo -e "${CYAN}ASH execution completed in ${TOTAL_EXECUTION} seconds.${NC}"
+    if [ "${DEBUG}" = "YES" ]; then
+      set -x
+    fi
+    echo "Resolved OCI_RUNNER to: ${RESOLVED_OCI_RUNNER}"
+
+    # Build the image if the --no-build flag is not set
+    if [ "${NO_BUILD}" = "NO" ]; then
```
orsifacundo, line 102: If you use the -q option then it takes some time between the message:
Resolved OCI_RUNNER to: xxxx
and the actual ASH execution.. so it seems that it hang, probably adding a message saying that build process is on the way or something like that. On other parts of ASH there are message telling you what it's actually being done

scrthq: Added

ash
scrthq: Done, good catch

scrthq:
> still looks OK to me. Approving.

Thank you, sir! I checked the aggregated results and wasn't seeing it in the scan list either.

begimher:
LGTM, awesome work!

begimher (approved):

Issue #46, opened 2024-03-13
feat(js): Add additional JS related extensions like .jsx, .ts, .tsx to extension filter for JS scan inclusion


scrthq:
Currently, ASH looks for `.js` files to determine if the JS scanners should be included. This is not always the case, e.g. when using exclusively Typescript or using `.jsx` instead of plain JS for frameworks like React.

Issue #44, opened 2024-03-07
fix(ash): Use .gitignore during archive expansion to ignore archives not included in the git index


scrthq:
When running ASH in a repository that includes archives already gitignored, those archives are unexpectedly expanded and scanned. This includes archives within 3rd party dependencies like node_modules and .venv directory trees.

ASH should treat these locations the same as it does during a scan and ignore them as candidates for expansion if the scan is being triggered from within a git repository.

awsmadi:
Fixed in v3.0.0 (PR #117, commit ba32f87). The v3 rewrite addressed this as part of the new Python-based architecture.

Issue #36, opened 2024-02-01
Mitigate potential risks of rm command


john-aws:
The [ash](https://github.com/awslabs/automated-security-helper/blob/main/ash) script contains two calls to `rm` of directory entries that are prefixed by shell variables:

```
rm -rf "${OUTPUT_DIR}"/work
rm -f "${OUTPUT_DIR}"/"${AGGREGATED_RESULTS_REPORT_FILENAME}"
```

While unlikely, it is not inconceivable that a bug could be introduced into the script whereby one or more of the shell variables evaluates to an empty string, thus causing `rm` to attempt to remove the wrong thing. The presence of the `-f` force flag adds additional jeopardy.

A famous variant of this problem was the [Ran steam. It deleted everything on system owned by user](https://github.com/ValveSoftware/steam-for-linux/issues/3671) bug where a Steam bash script ran `rm -rf "$STEAMROOT/"*` when `$STEAMROOT` was inadvertently empty. More discussion [here](https://www.theregister.com/2015/01/17/scary_code_of_the_week_steam_cleans_linux_pcs/).

Some suggestions for mitigating a mistake:

1. set -u
2. set -euo pipefail
3. check the shell variables are not empty before executing the `rm` command
4. so-called [bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/)

PR #47, opened 2024-03-13
added get-scan-set.py to utils scripts to return a list of non-ignored files for processing

scrthq:
*Issue #, if available:*
- #44 
- #46 
- #36

*Description of changes:*
- #44 - Adds new `get-scan-set.py` python script that allows extensible ignore lists to be provided (vs just using .gitignore). Does not require Git installed, uses Pathspec only in Python. **Uses this script to gather the list of non-ignored files to determine which zip files should be extracted.**
- #46 - Adds the additional extensions under JS_EXTENSIONS and GRYPE_EXTENSIONS known extension lists
- #36 - Guards `rm` calls with path evaluation 


begimher (approved):
LGTM

Issue #56, opened 2024-04-04
Support pnpm audit


bestickley:
I have a PNPM Monorepo. When I run ash on it, it uses `npm audit` which requires a package-lock.json which my project does not have. I have a pnpm-lock.yaml. Please support PNPM projects and use `pnpm audit`. Thank you!

awsntheule:
Hey @bestickley, ASH should now use pnpm when appropriate with PR #66. Please clone the latest changes and try again when you get a chance.

bestickley:
Thank you, @awsntheule, I'll try to get to this next week.

awsntheule:
@bestickley, have you had a chance to confirm your issue is fixed?

bestickley:
Confirmed this worked by seeing aggregated_results.txt:

>>>>>> Begin pnpm audit output for /src >>>>>>

No known vulnerabilities found

<<<<<< End pnpm audit output for /src <<<<<<

Thank you, @awsntheule !

PR #66, opened 2024-04-15
Add support for pnpm audit

awsntheule:
Issue #56 

Description of changes: This pull request adds support for the pnpm package manager's `pnpm audit` command.


scrthq:
@awsntheule I am wondering if we should potentially add `yarn` support here too? 🤔

awsntheule:
> @awsntheule I am wondering if we should potentially add `yarn` support here too? 🤔

That is a great idea! It seems doable to me. I'll get to work on it.

scrthq:
@awsntheule Awesome, thank you! It should be almost identical, just different lockfile name (yarn.lock). Yarn should be installable through `npm`.

awsntheule:
@scrthq I've added support for yarn with the latest commits. If you could take a look that would be great!

scrthq (approved):

Issue #72, opened 2024-04-24
v1.3.0 - grype and syft scanners are not running against items in git repositories


climbertjh2:
I have found that with v1.3.0, the `grype` and `syft` scanners are not scanning files that they should - but only when the `--source-dir` is set at a git repository.  If the `--source-dir` is not a git repository root, then the scanners find the files.

I have narrowed this down to the use of a `--tmpfs` mount for the container run.

climbertjh2:
I am working on a PR for this.

scrthq:
This looks good as of PR #73 ! Ready to close when you are done confirming from my perspective, I'm seeing Syft kicking in again where I wasn't previously

climbertjh2:
Yes - this is fixed now.  Closing.

PR #73, opened 2024-04-24
fix(ash): use /tmp rather than tmpfs for scratch area

climbertjh2:
*Issue #72 :*

*Description of changes:*

This updates the method used for using allocating and using temporary space.  The new method is to allocate a temporary folder/directory in `/tmp` within the container by using the `mktemp` command.  The old method, using a `--tmpfs` temporary file-system when launching the container, seems to cause the `grype` and `syft` tools to fail to find files that are placed into the temporary file-system (even though other file utilities like `ls` and `find` do see the files in that temporary file-system.


climbertjh2:
There is some additional updates required to clean up other uses of `/run/scan/src`.  However, these can be considered clean-up/re-factoring rather than fixing the current issue.  I suggest handling these other ones which are used when running multiple separate containers in parallel as a separate set of updates in a separate PR.

scrthq (approved):
Tested and am not seeing any regressions so far. Also validating that this improves Syft/Grype scanning as well! LGTM 🚀

PR #84, opened 2024-05-14
fix: update __version__ file to match release tag format in github.com

climbertjh2:
*Issue #, if available:* No number assigned

*Description of changes:*

Updated the reference to the public Github repository in `ash_multi` so that version checking works inside the container.

Also updated the `__version__` file to match the version numbering format in the tags in the Github repository.


rafaelpereyra (approved):
LGTM

PR #98, opened 2024-07-03
fix: Windows build issue due to CRLF on shell scripts

scrthq:
### What's Changed

- Fixed line endings on relevant files from CRLF to LF to resolve Windows build issues

### DIFF
--- a/utils/ash_helpers.ps1
+++ b/utils/ash_helpers.ps1
@@ -127,8 +127,7 @@ function Invoke-ASH {
                     '-t'
                     $AshImageName
                     "'$ashRoot'"
-                    $($buildArgs -join ' '
-                    '-q')
+                    $($buildArgs -join ' ')
                 ) -join ' '
                 $runCmd = @(
                     $RESOLVED_OCI_RUNNER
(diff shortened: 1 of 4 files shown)


utils/ash_helpers.ps1
```diff
@@ -127,8 +127,7 @@ function Invoke-ASH {
                     '-t'
                     $AshImageName
                     "'$ashRoot'"
-                    $($buildArgs -join ' '
-                    '-q')
+                    $($buildArgs -join ' ')
```
rafaelpereyra, line 130: Is this line needed?

rafaelpereyra (commented):
Small comment

utils/ash_helpers.ps1
scrthq: @rafaelpereyra yes, so that additional args collected for the build are passed into the `docker build` call. The reason for the change is I saw that the `-q` flag was being passed in every time for this build call, resulting in no build output at all regardless if -Quiet was included.

rafaelpereyra (approved):

climbertjh2:
Merging.

PR #104, opened 2024-08-12
Begin implementing support for offline mode

awsmadi:
*Issue #, if available:*

*Description of changes:*
Introduce support for executing ASH offline after building the container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,8 @@
 # Enable BASE_IMAGE as an overrideable ARG for proxy cache + private registry support
 #
 ARG BASE_IMAGE=public.ecr.aws/docker/library/python:3.10-bullseye
-ARG OFFLINE="NO"
-ARG OFFLINE_SEMGREP_RULESETS=""
+
+
 FROM ${BASE_IMAGE} as poetry-reqs
 
 ENV PYTHONDONTWRITEBYTECODE 1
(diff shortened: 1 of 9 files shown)


helper_dockerfiles/Dockerfile-grype
```diff
@@ -1,11 +1,24 @@
...
+ENV OFFLINE="${OFFLINE}"
+ENV OFFLINE_SEMGREP_RULESETS="${OFFLINE_SEMGREP_RULESETS}"
+ENV GRYPE_DB_CACHE_DIR="${HOME}/.grype"
 
 # Instal prerequisites
 RUN curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -b /usr/local/bin && \
     curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin && \
     python3 -m pip install semgrep
 
+RUN RUN if [[ "$OFFLINE" == "YES" ]]; then \
```
scrthq, line 15: Double `RUN` call here

scrthq: Also worth considering that this Dockerfile is going away soon; I think just correcting this one bit should be fine so it's syntactically correct, but I wouldn't spend too much more time on this one :-)

Dockerfile
```diff
@@ -4,6 +4,8 @@
 # Enable BASE_IMAGE as an overrideable ARG for proxy cache + private registry support
 #
 ARG BASE_IMAGE=public.ecr.aws/docker/library/python:3.10-bullseye
+ARG OFFLINE="NO"
+ARG OFFLINE_SEMGREP_RULESETS=""
```
scrthq, line 8: Should this also default to `"p/ci"` like the default in `ash`?

scrthq (changes requested):

scrthq (dismissed):
Tested and everything is working, actual scan is running noticeably faster as well thanks to the dependencies being included in the built image. Excellent job, @awsmadi !!!

scrthq (approved):
Looks good to merge from my perspective!

PR #162, opened 2025-07-10
fix(metrics): corrected suppression count issue and resolved overall metric inaccuracies

scrthq:


automated_security_helper/utils/sarif_utils.py
azharu, line 1: What's the rationale behind the multiple hasattr()?

Issue #184, opened 2025-08-22
uv cache error on all tools


rafaelpereyra:
Everytime ASH runs, there is a validation of the uv cache running the command `uv cache info`.

This command doesn't exist and causes this error:

```
2025-08-22 13:56:48,021	DEBUG	plugin_base.py:90	([yellow]jupyter[/yellow] @ [magenta]source[/magenta])	[INSTALLATION_CACHE] UV cache not available or has issues
Cache error: error: unrecognized subcommand 'info'

Usage: uv cache [OPTIONS] <COMMAND>

For more information, try '--help'. 
```

Then ASH continues without error. Given that the only commands possible for `uv cache` are:

```
Usage: uv cache [OPTIONS] <COMMAND>

Commands:
  clean  Clear the cache, removing all entries or those linked to specific packages
  prune  Prune all unreachable objects from the cache
  dir    Show the cache directory
```

I propose removing the code that causes this error.

cc @scrthq

PR #186, opened 2025-08-22
Fix/uvcache

rafaelpereyra:
*Issue #, if available:*
 #184
*Description of changes:*

Modified UV cache command from 'uv cache info' to 'uv cache dir' and simplified the cache information parsing logic to handle the single directory path output instead of searching for size/location keywords.


awsmadi (approved):
lgtm

PR #221, opened 2026-01-14
Fix/opengrep disabled default

awsntheule:
*Description of changes:* This change makes opengrep run by default on systems that are not using Windows. 


begimher (approved):
LGTM

Issue #223, opened 2026-01-23
--source-dir` argument ignored when using `ash scan` subcommand


prabnikaws:
# Bug: `--source-dir` argument ignored when using `ash scan` subcommand

## Summary

When running `ash scan --source-dir <path>`, the `--source-dir` argument was ignored and ASH scanned the current working directory instead.

## Root Cause

In `automated_security_helper/cli/scan.py`, the `run_ash_scan_cli_command` function is registered both as:

1. A typer callback (for running `ash` without a subcommand)
2. The `scan` subcommand (for running `ash scan`)

When a user runs `ash scan --source-dir tests/foo`, the function is invoked **twice**:

1. First as the callback with `ctx.invoked_subcommand="scan"` — using the **default** `source_dir` (current directory)
2. Then as the `scan` command with `ctx.invoked_subcommand=None` — using the **user-provided** `source_dir`

The original guard condition on line 296:

```python
if ctx.resilient_parsing or ctx.invoked_subcommand not in [None, "scan"]:
    return
```

This allowed the callback to execute when `invoked_subcommand="scan"`, causing the scan to run with the wrong `source_dir` before the actual `scan` command could run with the correct one.

## Fix

Change the condition to skip execution when any subcommand is invoked:

```python
if ctx.resilient_parsing or ctx.invoked_subcommand is not None:
    return
```

This ensures the callback only runs when no subcommand is specified (`ash --source-dir ...`), and the `scan` subcommand handles its own execution with the correct arguments.

## File Changed

`automated_security_helper/cli/scan.py` — line 296

### Compare behavior

Run these two commands that should be equivalent:

```bash
# Without subcommand (works correctly)
uv run ash --source-dir tests/test_data/scanners/bandit --scanners bandit --no-progress

# With scan subcommand (broken without fix)
uv run ash scan --source-dir tests/test_data/scanners/bandit --scanners bandit --no-progress
```

### Evidence

**Command 1: `ash --source-dir ...` (correct behavior)**

```
INFO     Found 2 files within the provided source directory to scan.
INFO     Scanner enablement validation summary:
INFO       Enabled scanners (1): bandit
INFO       Excluded scanners (5): cdk-nag, checkov, detect-secrets, npm-audit, semgrep
```

Only bandit runs, scanning 2 files in the target directory.

**Command 2: `ash scan --source-dir ...` (buggy behavior)**

```
INFO   Processing 131 SARIF results for OCSF report
[...]

PR #224, opened 2026-01-23
fix: respect --source-dir argument in ash scan subcommand

prabnikaws:
*Issue #, if available:* #223

*Description of changes:*

Fixes a bug where `--source-dir` and other CLI arguments were ignored when using the `ash scan` subcommand.

Changed the condition in `scan.py` line 296 from:

`if ctx.resilient_parsing or ctx.invoked_subcommand not in [None, "scan"]:`

to:

`if ctx.resilient_parsing or ctx.invoked_subcommand is not None:`

Also removed stray debug print statement from `run_ash_scan.py`.

*Files changed:*
- `automated_security_helper/cli/scan.py` - Fixed callback guard condition
- `automated_security_helper/interactions/run_ash_scan.py` - Removed debug print statement

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ ASH v3 integrates multiple open-source security tools as scanners. Tools like Ba
 curl -sSfL https://astral.sh/uv/install.sh | sh
 
 # Create an alias for ASH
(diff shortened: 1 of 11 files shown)


rafaelpereyra:
Merging since the issue seems to be with Snyk on the pipeline (token issues)

PR #279, opened 2026-04-28
fix: config file discovery broken since v3.2.7

awsmadi:
## Summary

Two bugs introduced in PR #274 that broke `.ash.yaml` config file discovery:

1. **`resolve_config()` skips file search when `source_dir` is `None`** — PR #274 changed the `source_dir` default from `Path.cwd()` to `None` and added an early return at line 172. The orchestrator never passes `source_dir`, so config discovery was completely bypassed.

2. **`.ash.yaml` has invalid `ignore_paths` format** — entries were bare strings (`"tests/test_data/**"`) but the `IgnorePathWithReason` model requires `{path: ..., reason: ...}` objects. Config validation failed silently and fell back to defaults.

## Fix

- Pass `self.source_dir` from the orchestrator to `resolve_config()`
- Fix `.ash.yaml` to use the correct `{path, reason}` format

## Test plan

- [ ] `uv run python -c "from automated_security_helper.config.resolve_config import resolve_config; from pathlib import Path; c = resolve_config(source_dir=Path('.')); print(c.global_settings.ignore_paths)"` shows the ignore path
- [ ] Run ASH scan on the repo — confirm suppressions are applied

### DIFF
--- a/.ash.yaml
+++ b/.ash.yaml
@@ -2,7 +2,8 @@ global_settings:
   # Ignore test data that intentionally contains insecure patterns
   # (these are scanner test fixtures, not production code)
   ignore_paths:
-    - "tests/test_data/**"
+    - path: "tests/test_data/**"
+      reason: "Scanner test fixtures with intentional insecure patterns"
 
   suppressions:
     # Test files for secret masking contain intentional fake passwords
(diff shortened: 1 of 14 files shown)


rafaelpereyra (approved):
Let's switch back the version to v3.3.1

PR #286, opened 2026-05-05
feat: config lint command

rafaelpereyra:
## Summary

- Add `ash config lint` command with auto-fix for config issues (internal fields, missing line_end, expired/unused suppressions)
- Add schema regeneration and smoke test to the release workflow
- Fix `mcp-resource-management` treated as invalid user config field
- Wrap all network-dependent Dockerfile installs with `with-retry` (3 attempts, exponential backoff) to handle transient CDN flakiness

## Dockerfile reliability changes

- Wrap curl-based installs: uv, pip, yarn, pnpm, syft, grype, trivy, nodesource GPG key, semgrep rulesets
- Wrap package manager installs: gem (cfn-nag), npm
- Add `set -o pipefail` to `with-retry.sh` so pipe failures propagate
- Fix GPG key pipe placed inside retry scope (prevents partial-data corruption)
- Fix OFFLINE semgrep loop quoting (pre-compute outfile to avoid nested escape cascade)
- Fix `assets/Dockerfile` COPY paths for pip-installed build context

## Test plan

- [x] `ash config lint` command exists and shows help
- [x] `ash config lint` catches config issues (missing required fields, unknown fields)
- [x] `ash config lint --fix --non-interactive` runs without error
- [x] Unit tests pass (255 pass, 1 pre-existing unrelated failure in mcp debug prints)
- [x] `with-retry.sh` syntax valid, `set -o pipefail` present
- [x] OFFLINE semgrep loop produces correct filenames (ci.yml, security-audit.yml, python.yml)
- [x] Dockerfile passes `docker build --check` syntax validation
- [x] `assets/Dockerfile` uses relative COPY path, root uses full path
- [x] GPG key pipe is inside retry scope in both Dockerfiles
- [ ] CI container build jobs pass (yarnpkg.com CDN dependent)
- [ ] OFFLINE=YES Docker build succeeds end-to-end

### DIFF
--- a/.github/workflows/ash-create-release.yml
+++ b/.github/workflows/ash-create-release.yml
@@ -62,6 +62,17 @@ jobs:
         if: steps.bump.outputs.bumped == 'true'
         run: uv run python scripts/version_template_manager.py generate
 
+      - name: Regenerate JSON schemas
+        if: steps.bump.outputs.bumped == 'true'
+        run: uv run python -m automated_security_helper.schemas.generate_schemas
+
+      - name: Smoke test release artifacts
+        if: steps.bump.outputs.bumped == 'true'
+        run: |
+          uv run ash --help
+          uv run ash config validate
+          uv run python -m pytest tests/unit/cli/test_config_init.py -x -q --no-header
+
       - name: Create release PR
         if: steps.bump.outputs.bumped == 'true'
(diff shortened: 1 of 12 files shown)


awsmadi (approved):
lgtm

PR #302, opened 2026-05-08
fix(ci): skip PR title check in merge queue context

awsmadi:
## Summary

- Skip the conventional commit title regex when triggered by `merge_group` events
- `github.event.pull_request.title` is null in merge queue context, causing the check to always fail
- The title was already validated on the original `pull_request` open/edit event

## Root cause

The workflow declares `merge_group:` as a trigger (required for merge queue status checks), but the validation step reads `github.event.pull_request.title` which doesn't exist in `merge_group` events. Empty string fails the regex.

## Test plan

- [x] Verified regex passes all current PR titles locally
- [ ] Merge this PR via merge queue to confirm the fix

### DIFF
--- a/.github/workflows/ash-pr-title-check.yml
+++ b/.github/workflows/ash-pr-title-check.yml
@@ -15,21 +15,22 @@ jobs:
     name: Validate PR title
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


awsmadi:
breaking glass to fix the merge queue

PR #291, opened 2026-05-06
feat: add GitHub Advanced Security (GHAS) SARIF reporter

rafaelpereyra:
Add a dedicated reporter that produces SARIF optimized for GitHub Code Scanning. Addresses the issue where all findings display as 'Error' instead of proper severity levels (Critical/High/Medium/Low).

Key changes:
- New github_ghas_reporter.py that ensures every rule has properties['security-severity'] set (required by GitHub)
- Flattens rules from tool.extensions into tool.driver.rules
- Strips verbose fields for ~80% size reduction vs standard SARIF
- Excludes suppressed findings by default (configurable)
- Preserves scanner-provided security-severity when present
- Falls back to deriving severity from defaultConfiguration.level or result.level when missing
- Fix generate_schemas.py to only write when content changes

Output: reports/ash.ghas.sarif (enabled by default)

Configuration:
  reporters: github-ghas: enabled: true options: exclude_suppressed: true

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #303, opened 2026-05-08
fix(ci): skip SARIF upload in merge_group events

awsmadi:
## Summary

- Skip `github/codeql-action/upload-sarif` when event is `merge_group`
- The `security-events: write` permission required by upload-sarif is not available in merge queue context
- SARIF upload runs on `push` to main after merge instead

## Root cause

Merge queue creates a temporary branch (`gh-readonly-queue/main/...`) where the GITHUB_TOKEN lacks `security-events: write`. The upload-sarif action fails with "Resource not accessible by integration."

## Test plan

- [ ] PR #291 enters merge queue without the SARIF upload failure


awsmadi:
breaking glass to fix merge queue

PR #333, opened 2026-05-12
refactor(cli): convert cli/mcp.py to a package — precondition for downstream MCP work

awsmadi:
## Summary
Single-file rename: `automated_security_helper/cli/mcp.py` → `automated_security_helper/cli/mcp/__init__.py` (91% similarity per git, with a 6-line docstring addition explaining the change).

This is a **precondition PR** — it disambiguates the package shape so downstream PRs can introduce sibling modules under `cli/mcp/` (progress_monitor, mcp_server, source_delivery, sessions, profile_registry, etc.) without dragging in the full streamable-HTTP transport feature for the package shape to be valid.

## Why this PR exists
The full refactor sequence introduces sibling modules under `cli/mcp/` in one commit and moves the legacy `mcp_command` symbols into `cli/mcp/__init__.py` in a much later commit. Splitting at any point in between leaves Python with a package (`cli/mcp/`) that **shadows** the legacy module file (`cli/mcp.py`) — `tests/unit/cli/test_mcp_cli.py` fails to import \`validate_log_options\` because the empty `__init__.py` resolves first.

This PR moves the legacy symbols into `cli/mcp/__init__.py` *before* any sibling modules are added, so subsequent PRs land cleanly.

## Confirmed NOT included
- streamable-HTTP transport, auth middleware, CORS handling, uvicorn, Starlette
- New typer flags (\`--transport\`, \`--host\`, \`--port\`, \`--mount-path\`, \`--auth-header-name\`, \`--auth-header-value\`)
- New dependencies (no \`pyproject.toml\` change)
- MCP profile registry, session lifecycle, source delivery
- Per-result scanner_details / template-method scanner refactor

Those land in their own PRs after this one merges.

## Test plan
- [ ] CI: \`tests/unit/cli/test_mcp_cli.py\` 29/29 (unchanged from main baseline)
- [ ] CI: full unit suite collects 2008 tests with 0 errors
- [ ] CI: no integration regressions
- [ ] Manual: \`from automated_security_helper.cli.mcp import mcp_command, validate_log_options\` works

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,77 @@
+name: Bug Report
+description: Report a bug in ASH
+labels: ["bug"]
+body:
+  - type: dropdown
+    id: version
+    attributes:
+      label: ASH Version
+      description: Which version of ASH are you using?
+      options:
+        - v3.4.x
+        - v3.3.x
+        - Other
+    validations:
+      required: true
+
+  - type: dropdown
+    id: os
+    attributes:
+      label: Operating System
+      options:
+        - Linux
+        - macOS
+        - Windows
+    validations:
+      required: true
+
+  - type: dropdown
+    id: install-method
+    attributes:
+      label: Installation Method
+      options:
+        - uvx
+        - pip
+        - pipx
+        - container
+        - homebrew
+    validations:
+      required: true
+
+  - type: textarea
+    id: description
+    attributes:
+      label: Bug Description
+      description: A clear description of the bug.
+    validations:
+      required: true
+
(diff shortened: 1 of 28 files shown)


awsmadi:
Closing in favor of the rebuilt PR 1 (`pr/scan-decomposition`) which inlines the `cli/mcp.py` → `cli/mcp/__init__.py` rename as part of its scope. The package-shape disambiguation needs to live alongside the `bf476a7` mcp_server.py split (which introduces the empty package shadowing) rather than as a separate precondition PR. The base-branch dependency graph (5 transitional model-method commits between current main and the refactor wave's working baseline) made the standalone-precondition approach require redundant rebase work.

PR #69, opened 2024-04-19
fix(cdk-nag-scan): copy output files to separate folders

climbertjh2:
*Issue #, if available:* 68

*Description of changes:* Updated `cdk-docker-execute.sh` to copy template/nag files on every iteration through the loop, copying to separate folders for each file scanned.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #338, opened 2026-05-12
refactor(scanners): template method + Track 2.1 migrations (#69)

awsmadi:
## Summary
Establishes the scanner template-method pattern in \`ScannerPluginBase\` and migrates built-in scanners to use it. Depends on **PR #334 (scan-decomposition)**.

### What's new
- **\`ScannerPluginBase.scan\`** is now a template method with hooks: \`_execute_scan\`, \`_invocation_extras\`, \`_read_results_file\`, \`_handle_empty_results\`, \`_post_process_sarif\`, \`_inject_invocation\`, \`_ensure_runs\`. Subclasses override only the unique scanner logic.
- \`_inject_invocation\` populates per-result and tool-driver \`scanner_details.tool_invocation\` (closes DA r7 #3 metadata regression).
- \`_execute_scan\` is \`@abstractmethod\`; every \`ScannerPluginBase\` subclass that overrides \`scan()\` directly provides a stub raising \`NotImplementedError\` to satisfy the contract.
- \`shlex.join\` for SARIF \`commandLine\` so shell-special characters round-trip correctly.
- \`uv_tool_runner\`: thread-safe cache with 5s probe timeout, per-key probe locks via WeakValueDictionary (avoids serializing parallel cold-cache scanner probes).

### Migrated scanners
- \`checkov\` and \`grype\` (initial migration)
- \`semgrep\` + \`opengrep\` (consolidated into shared \`GrepScannerBase\`, then base migrated to template)
- \`bandit\` (override \`success_exit_codes={0,1}\` for findings-present)
- **Deferred** (genuinely-bespoke or Python-API-based): \`cdk_nag\`, \`cfn_nag\`, \`detect_secrets\`, \`npm_audit\`, \`syft\`, \`snyk_code\`, \`trivy_repo\`, \`ferret-scan\` — all carry the \`_execute_scan\` stub and keep their existing scan() override.

### Commits (14)
Includes 13 cherry-picks plus one partial extraction of DA r7 #3 (per-result \`scanner_details\` attach in template).

## Dependencies
- **Required**: #334 (scan-decomposition)

## Test plan
- [ ] CI: \`tests/unit/base/test_scanner_template_method.py\` (new)
- [ ] CI: \`tests/unit/plugin_modules/\` — 473/474 passing (1 pre-existing failure on the base, unrelated to this PR)
- [ ] CI: full unit suite collects 2567 tests, 0 errors


rafaelpereyra (approved):
Does this refactor should also update the documentation on building plugins? Could be a docs sweep just wanted to be sure is included @awsmadi

rafaelpereyra (dismissed):

awsmadi:
@rafaelpereyra good catch — yes, and it was more than a docs sweep.

The refactor makes `_execute_scan()` the sole abstract method on `ScannerPluginBase`, with `scan()` becoming a template method around it. That's a breaking change for third-party scanner plugins: a subclass that overrides `scan()` and doesn't declare `_execute_scan()` is still abstract and fails at instantiation. This PR already had to add exactly that stub to `examples/ash_plugins_example/my_ash_plugins/scanner.py`, which was the tell that the published guidance needed updating too.

Three docs asserted the old contract and are fixed in e0dda1c:

- **`scanner-plugins.md`** — every example overrode `scan()` with no `_execute_scan`, so as written none of them could be instantiated. Now documents the template flow, an `_execute_scan` example returning `(argv, results_file, subprocess_env)`, the adjustable hooks, and the override-`scan()`-directly variant with its required stub.
- **`development-guide.md`** — the interface block declared `@abstractmethod scan()` and `@abstractmethod cleanup()`. `scan()` isn't abstract and `cleanup()` doesn't exist at all.
- **`scanner-plugins-diagrams.md`** — the class diagram advertised `scan()` as the subclass extension point and returned `ScanResultsContainer`.

Left alone after checking: `plugin-best-practices.md` subclasses `ScannerPluginBase` only to illustrate the `dependencies` field, so it makes no claim about the scan contract, and `scanners-diagrams.md` / `trivy-plugin.md` have no contract references.

Claims were verified against the loaded class rather than the diff — `__abstractmethods__ == {'_execute_scan'}`, `hasattr(..., 'cleanup')` is `False`, and the documented signature matches `inspect.signature`. Docs freshness is 7/7.

rafaelpereyra (approved):

PR #368, opened 2026-06-18
fix: update vulnerable dependencies and switch to slim base image

rafaelpereyra:
## Summary

- **uv.lock**: Upgrade 4 packages with known CVEs (cryptography, pyjwt, python-multipart, starlette)
- **Dockerfile**: Switch base from `python:3.12-bookworm` to `python:3.12-slim-bookworm`

## Trivy Scan Results

| Metric | Before | After | Improvement |
|---|---|---|---|
| Image size | 3.66 GB | 3.09 GB | -16% |
| OS CRITICAL | 41 | 20 | -51% |
| OS HIGH | 343 | 235 | -31% |
| Total CRITICAL+HIGH | 384 | 255 | -34% |

## What Changed in Dockerfile

- Base image: `python:3.12-bookworm` → `python:3.12-slim-bookworm`
- Added `build-essential`, `ca-certificates`, `gnupg` to core stage (needed for Ruby gem compilation in slim)
- Consolidated duplicate `apt-get update` + package installs from Node.js block into main apt block

## Remaining CRITICAL Sources (not addressable via base image change)

- `ruby3.1` (Debian-packaged, required by cfn-nag)
- `perl` (pulled by build-essential for native extensions)

## Testing

- `docker build --target core` ✅
- `ash --version` inside container → v3.5.4 ✅
- All scanners install successfully ✅


prabnikaws (dismissed):
Approving. Tightly scoped security fix: - uv.lock: bumps CVE-affected deps (cryptography 48.0.0->49.0.0, pyjwt 2.12.1->2.13.0, python-multipart 0.0.28->0.0.32, starlette 1.3.0->1.3.1; pydantic-settings 2.14.1->2.14.2 via lock resolution). All real PyPI releases. - Dockerfile: switch to python:3.12-slim-bookworm, with build-essential/ca-certificates/gnupg added to compensate and the redundant Node.js apt block consolidated. Correct and consistent. - .ash_community_plugins.yaml: removes an expired (2026-06-01) suppression and ignores CI-generated test-results/**. Checks passing and ASH self-scan reports no actionable findings. Remaining CRITICALs (ruby3.1, perl) are acknowledged and not addressable via base image. LGTM.

nikmeiser (approved):
Non-blocking follow-up notes from review (the change itself looks good and is well-covered by the end-to-end container scan jobs):

1. build-essential lands in the runtime image. It's installed in the core stage, which both ci and non-root inherit, so gcc/make/dpkg-dev (+ perl) ship in the final image. It's needed to compile cfn-nag's Ruby native gems at build time, but likely not at runtime. Confining it to a builder stage and copying the built gems would reclaim more of the slim savings (probably why the size delta is ~-16% rather than larger). Non-blocking optimization for a follow-up.
2. cryptography 49.0.0 is a major bump with backwards-incompatible changes. Notably it drops prebuilt x86_64 macOS and 32-bit Windows wheels. No impact on the Linux container, and the full unit-test matrix passes — but worth a heads-up that local pip/uvx installs on Intel Macs will now build from source (or fail without a toolchain), which CI won't catch since the macOS runners are arm64.
3. Documentation note (out of scope for this PR, flagging for follow-up): no doc changes are needed here — nothing in docs/ or the README references the base image, image size, or supported architectures, so this change doesn't make any existing docs stale. That said, there's a pre-existing gap worth a separate issue: the docs don't have a supported platforms / architectures section. With cryptography 49 dropping prebuilt x86_64 macOS and 32-bit Windows wheels, a short note in the installation guide clarifying which OS/arch combinations ship prebuilt wheels (vs. build-from-source) would help local pip/uvx installers set expectations. Not a blocker for this PR.

PR #386, opened 2026-07-02
build(docker): stop recursive chmod/chown from copying up the Grype DB (~1.7GB / 30% smaller image)

awsgoffy:
## Summary
The offline container image is ~1.7 GB larger than necessary. In the `core`
stage a trailing `chmod -R … /deps` runs *after* `/deps/.grype` has been filled
with the offline Grype vulnerability database (~1.7 GB). On an overlay
filesystem, changing a file's mode triggers a **copy-up** of that file into the
new layer, so the recursive chmod rewrites the entire DB into a fresh ~1.7 GB
layer. The `non-root` stage repeats this with `chown -R … /deps`.

## Fix
Set permissions **without a separate recursive pass over the populated DB**:
1. `chmod` the Grype/Semgrep/opengrep cache dirs **at creation**, while empty.
2. In the OFFLINE block, `chmod -R` the bundled DB + rules **in the same RUN that
   downloads them** — files created and chmod'd in one layer don't copy up, so
   the DB is counted once, yet stays world-readable for arbitrary runtime UIDs.
3. Drop `/deps` from the core-stage trailing `chmod` and the non-root
   `chown -R`/`chmod -R`.

See `Dockerfile.patch` (generated from the actual branch diff).

## Verified (built and tested, not just reasoned)
Built with `OFFLINE=YES` before and after the change:

| target | before | after | reduction |
|---|---|---|---|
| `ci` | 5.788 GB | **4.045 GB** | −1.74 GB (~30%) |
| `non-root` | 8.191 GB | **4.705 GB** | −3.49 GB (~43%) |

(`non-root` benefits twice: it previously duplicated the DB in both the core
`chmod -R` and its own `chown -R`.) The duplicate chmod layer is gone; the
~1.741 GB Grype DB layer is unchanged, as expected.

**Functional parity:**
- `ci` run as a random UID (`--user 4321:0`) with `--network none`: Grype read
  the bundled DB and reported ~79 CVEs (incl. Log4Shell).
- `non-root` run as its default non-root user with `--network none`: same — DB
  read successfully offline.

Confirms the arbitrary/non-root UID read access that the original
`chmod -R 777 /deps` provided is preserved.

> Note: an earlier attempt that only chmod'd the dirs at creation (without step 2)
> shrank the image but **broke** arbitrary-UID DB access — `grype db update`
> creates the DB files *after* that chmod. Step 2 is what keeps parity. Worth
> calling out so reviewers don't "simplify" it back.

## Trade-off to flag for maintainers
Dropping `/deps` from the `non-root` `chown -R … && chmod 750 -R …` leaves `/deps`
world-accessible (dirs 777 / files 644) rather than user-owned 750. If you prefer
`/deps` not world-writable, an alternative that still avoids the copy-up is to
create the cache dirs with a shared group (e.g. GID 100, mode 775) in `core` and
rely on group membership. Happy to switch to whichever the team prefers.

## Test plan (to reproduce)
```bash
finch build --target ci --build-arg OFFLINE=YES \
  --build-arg OFFLINE_SEMGREP_RULESETS="p/java p/security-audit p/secrets" \
  -t ash:ci-after .
finch images | grep ash          # compare size vs baseline
finch history --no-trunc ash:ci-after | grep -i 'chmod -R 755'   # layer now tiny
# arbitrary-UID, offline DB read:
finch run --rm --network none --user 4321:0 \
  -e GRYPE_DB_AUTO_UPDATE=false -e GRYPE_DB_CACHE_DIR=/deps/.grype \
  -v ./sample:/src:ro ash:ci-after grype dir:/src -o table   # expect CVEs
```


awsmadi (approved):
Thank you for your contribution!

awsgoffy:
Thanks @awsmadi. Is there anything more you need from me on this, or is it just waiting on your side?

PR #410, opened 2026-08-13
fix(docs): restore working star history charts

Dessalines39394:
The star history chart in the README and docs is currently broken because it depends on the GitHub stargazer API, which has imposed restrictions that prevent the chart from loading. This patch points the chart at an alternative provider that uses the same domain for both the API and frontend and requires no API token, so the chart renders correctly again. Updates both the generated documentation and the corresponding templates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,4 +483,4 @@ This library is licensed under the Apache 2.0 License. See the [LICENSE](LICENSE
 
 ## Star History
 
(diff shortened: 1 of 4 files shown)


awsmadi:
We appreciate your contribution but will be moving to an Actions-based approach. Thank you!

PR #456, opened 2026-08-25
feat: severity ladder, path containment, and workspace path conversion (workspace mode phase 0)

awsmadi:
First of six PRs implementing workspace mode (`--workspace`), which scans N project directories as N independently-scoped executions and aggregates them with per-project attribution.

This PR ships **no user-visible feature**. It adds the shared primitives the later phases need, plus two independent bug fixes that were found while building them. Later PRs in the stack: workspace resolution and `--dry-run`, scoped execution and aggregation, the reporter dimension, workspace-level config, and MCP integration.

## Why these primitives

Workspace mode has exactly two path spaces: project-relative inside a project's execution (byte-for-byte today's behaviour), and workspace-relative in aggregated output only. Conversion happens in one specified place rather than by string concatenation. Suppressions are never rewritten — a project's suppressions are matched inside that project's execution, which deletes a class of bug rather than specifying a workaround for it. Workspace-level suppressions go the other way: they are declared workspace-relative and pushed *down* into each project, and are passed to a project only if they can match inside it.

## What is here

| File | Purpose |
| --- | --- |
| `utils/severity_ladder.py` | The severity-threshold strictness ladder, encoded once |
| `utils/path_containment.py` | Canonicalise, reject outside-root, reject symlinked entries, reject `..` |
| `utils/workspace_paths.py` | Pattern conversion between the two path spaces, both directions |
| `models/workspace.py` | Exit-code contract and the `skipped_projects` payload |

## The threshold direction is counter-intuitive, so it is stated explicitly

In ASH, **raising** a severity threshold **loosens** the gate. `determine_status` fails on `high` when the threshold is any of `ALL`/`LOW`/`MEDIUM`/`HIGH`, but on `medium` only for `ALL`/`LOW`/`MEDIUM`. So the strictness order is:

```
ALL (strictest)  <  LOW  <  MEDIUM  <  HIGH  <  CRITICAL (loosest)
```

CRITICAL findings fail regardless of any threshold. An empty or `None` threshold behaves as the most permissive value, which the refactor preserves.

This matters because the workspace-level field in a later PR is `max_severity_threshold` — the loosest a project is allowed to be. A field named for the minimum would tighten and loosen in the wrong directions.

## The `determine_status` refactor is behaviour-preserving, and that is verified

`ScanResultsContainer.determine_status` was a five-branch cascade; it now delegates to the ladder. Equivalence was checked by transcribing the original cascade and brute-forcing every threshold value (including `None`, `""`, and an unrecognised string) against 243 severity-count vectors — 3,888 pairs, zero mismatches. The check carries its own vacuity guard, because zero mismatches is also what you get if the counters never land: the old-verdict distribution came out 2,390 FAILED / 1,498 PASSED, so both arms were genuinely exercised.

The committed test keeps its oracle in the original membership-test shape while the implementation compares integer ranks, so it is a genuine cross-check rather than a restatement of the new code.

## Two bug fixes, both independent of workspace mode

**junitxml ignored the `LOW` and `ALL` thresholds.** `junitxml_reporter.py` decided actionability from SARIF levels with branches for only `CRITICAL`, `HIGH` and `MEDIUM`; `LOW` and `ALL` fell through leaving every finding actionable. Now routed through the shared ladder. Exactly one cell changes among the five recognised thresholds (`LOW` + `none`, actionable to not-actionable), and a malformed threshold string now gates on critical only instead of marking everything actionable. Reachable only for externally-supplied SARIF: nothing in `automated_security_helper/` writes `properties.severity_threshold`.

**A comment in `suppression_matcher.py` had the wildcard semantics backwards.** It claimed `fnmatch` treats `*` as "anything except `/`". It does not — `fnmatch.translate("*")` is `(?s:.*)`, so `*` crosses `/`. The practical consequence is the opposite of what the comment implied: a pattern like `src/*.py` also matches `src/sub/deep/a.py`, so such a suppression silently covers subdirectories. Behaviour unchanged; only the comment.

## Down-conversion fails closed rather than guessing

Pushing a workspace-level pattern down into a project cannot always be done soundly, because `file_path_matches` has two different glob semantics: patterns without `**` go to `fnmatch`, where `*` crosses `/`, while patterns containing `**` go to `_recursive_glob_match`, which anchors per path component. Rewriting a leading `*` as `**` therefore re-interprets every other glob in the pattern.

So `to_project_pattern` converts only the shapes proven to preserve which files a pattern matches, returns "cannot apply here" when the pattern provably cannot match inside a project, and raises otherwise, naming the pattern and the project. The contract it holds is:

```
for every project-relative path p:
    matches(to_project_pattern(P, R), p) == matches(P, R + "/" + p)
```

That contract is property-tested, and was independently swept over 50 patterns x 4 prefixes x 11 paths (1,375 triples) with zero violations across all three outcomes. The fail-closed classes are not merely declared: each test exhibits a path where every candidate rewrite gets the answer wrong, so raising is the only honest option.

Getting this wrong is security-relevant in one direction specifically. A suppression that under-applies leaves an extra finding in the report, which is visible. A suppression that over-applies silences a real finding, which is not.

## Verification

- Full unit suite: 3546 passed, 22 skipped, 0 failures, under both `-n auto` and serial. Baseline on `main` is 3015 passed.
- `ruff check` clean. `ruff format` reports only `suppression_matcher.py`, which is pre-existing on `main` in the regex block near line 22 and untouched here.
- `scripts/verify_docs_freshness.py` 7/7.
- All commit messages pass `cz check`.

Known limitation: the contract property test skips the fail-closed pattern classes and defers them to `TestFailClosedIsNecessary`, which enumerates its cases explicitly rather than deriving them from the same matrix, so the two sets are not mechanically tied together.


awsmadi:
Addressed the eight code-scanning comments. Bandit run locally with this repo's own `.ash/bandit.yaml` reports **six of the eight as live**; the other two do not fire.

**The six:**

- `run_ash_scan.py:567` (B101) — `assert opts.workspace_plan is not None` replaced with an explicit raise. `python -O` strips assert, so under `-O` a missing plan would have surfaced later inside `execute_workspace` as an `AttributeError` on `None`. `RuntimeError` rather than a `Workspace*Error`, which would report a caller bug as a problem with the operator's workspace file. The branch is unreachable through the CLI (the sole caller already guards it), so it now has a test — mutation-checked by reverting the raise, which fails both new tests.
- `verify_workspace_reports.py:91` (B404) — added the `# nosec B404` comment its four sibling gate scripts already carry. This file already uses `# nosec B603` on line 223, so it was an inconsistency rather than a new judgment.
- `verify_docs_freshness.py` (B112) — the broad catch-and-continue when a scanner module will not import is deliberate and documented in place; a missing optional extra is not a docs problem. Suppressed with a reason, matching the sixteen `# nosec B110` suppressions already in the tree.
- Three B105 "possible hardcoded password" (×2 `test_workspace_container_mode.py`, ×1 `test_multi_project_attribution_gate.py`) — false positives on `if token == "--mount"` / `"--scanners"`. Bandit flags a literal compared against an identifier whose name looks secret-ish, and `token` is on that list. Renamed to `arg`, removing the finding at its source rather than suppressing it.

**The two that do not fire:** the B108 comments on `test_path_containment.py:271` and `test_policy_wiring.py:158` both point at *comments explaining that the code deliberately avoids `/tmp`*. Bandit reports nothing at either location. Alerts are anchored on (rule, path, line), so a line shift re-anchors an old alert onto whatever now occupies that line. Both files already handle B108.

**Verification**

```
bandit -c .ash/bandit.yaml -r .     44 findings -> 38
in the eight alerted files           6 -> 0
by rule    B404 8->7, B101 3->2, B105 5->2, B112 1->0
```

Full unit suite 4516 passed, 0 failed. Schemas clean, docs freshness 8/8. ruff on the touched files is at the same 81 pre-existing findings as before, and the one file ruff would reformat needed it already, in code this commit does not touch.

rafaelpereyra (approved):

PR #457, opened 2026-08-25
fix(config): describe what a suppression without line_end actually matches

### LINKED ISSUE #347: Suppression with `line_start` but no `line_end` is matched as open-ended, but lint message claims it defaults to `line_start`
## Summary
There is an inconsistency between what `ash config lint` tells the user about a suppression with only `line_start` set, and what `_line_range_matches` actually does at scan time.

- **Lint message** (`config_linter.py::_check_suppression_issues`):
  > "Suppression has 'line_start' (N) but missing 'line_end' — will default to line_start value"
- **Matcher** (`suppression_matcher.py::_line_range_matches`) treats `line_end=None` as **open-ended** ("from `line_start` to end of file"), not as `line_end=line_start`.

The matcher and the suppression-id generator also disagree with each other:
- `_line_range_matches` (matching) treats missing `line_end` as `+∞`.
- `_get_suppression_id` (in both `sarif_utils.py` and `unused_suppressions_reporter.py`) substitutes `line_start` for missing `line_end` when building the dedupe key.

## Why it matters
- A user sets `line_start: 42` expecting "only line 42", but every finding on line ≥ 42 in that file matches. This silently over-suppresses.
- The lint warning gives users false confidence about the runtime semantics.

## Reproduction
```yaml
global_settings:
  suppressions:
    - path: "src/foo.py"
      rule_id: "B201"
      line_start: 42
      reason: "False positive"
```
A finding at `src/foo.py:99` with `rule_id=B201` is suppressed, even though the user only specified line 42.

## Relevant code
- `automated_security_helper/utils/suppression_matcher.py::_line_range_matches` — branch `if suppression.line_start is not None and suppression.line_end is None:` returns true for `finding_end >= suppression.line_start`.
- `automated_security_helper/config/config_linter.py::_check_suppression_issues` — emits the misleading "will default to line_start value" message.
- `automated_security_helper/utils/sarif_utils.py::_get_suppression_id` and `automated_security_helper/plugin_modules/ash_builtin/reporters/unused_suppressions_reporter.py::_get_suppression_id` — substitute `line_start` for missing `line_end` in the ID.
- `automated_security_helper/models/core.py::AshSuppression` — `line_end` description is just `(Optional) Ending line number`; doesn't document missing-value semantics.

## Proposed fix (pick one and align everything)

**Option A — make missing `line_end` mean "single line"** (matches lint message, matches user intuition, matches `_get_suppression_id`):
- Change `_line_range_matches` to default `line_end = line_start` when missing.
- Update `AshSuppression.line_end` docstring to say so.
- Keep current lint message and `_get_suppression_id` behavior.

**Option B — keep open-ended semantics** (current matcher behavior):
- Update `_line_range_matches` docstring to document open-ended behavior.
- Update lint message in `_check_suppression_issues` to:
  > "Suppression has 'line_start' (N) but missing 'line_end' — will match all findings from line N to end of file. Set line_end explicitly."
- Update `_get_suppression_id` in both `sarif_utils.py` and `unused_suppressions_reporter.py` to use a sentinel like `*` (or `EOF`) instead of `line_start` for missing `line_end`, so unused-suppression dedupe matches the actual semantics.
- Update `AshSuppression.line_end` docstring.

Option A is the lower-risk choice; Option B is a behavior change for anyone relying on the current open-ended behavior.

## Acceptance criteria
- `ash config lint` message and the runtime matcher behavior agree.
- `_get_suppression_id` is consistent across `sarif_utils.py` and `unused_suppressions_reporter.py` and reflects the chosen semantics.
- `AshSuppression.line_end` docstring documents the semantics of the missing case.
- Tests in `tests/unit/utils/test_suppression_matcher_extended.py` cover the missing-`line_end` case explicitly.

awsmadi:
## The contradiction

`_line_range_matches` with no `line_end`:

```python
return finding_end >= suppression.line_start
```

That matches **every finding from that line to the end of the file**. The linter said the opposite:

```
Suppression has 'line_start' (40) but missing 'line_end' - will default to line_start value
```

Whichever one you believed, the other was lying. Someone writing `line_start: 40` to silence one finding was told they had done exactly that, while actually suppressing everything from line 40 onwards — including findings that did not exist yet.

## Message corrected, matcher left alone

Narrowing the matcher is the other way to reconcile them, and it is **breaking**: every existing single-line suppression would stop covering the range it covers today, so findings currently suppressed reappear as actionable on upgrade. Rewording costs nothing and breaks nobody, so open-ended matching becomes the documented contract.

That makes it a decision rather than an accident, which is why the tests **pin the matcher too**. Without them the next reader of `finding_end >= line_start` has no way to tell whether it is intended — which is how this got out of step in the first place. Those three tests pass on `main`; they are there to stop a future "tidy-up" from silently narrowing it.

## The autofix is now described as a narrowing

`Set line_end = 40 (same as line_start)` reads as correcting an omission. It is not: applying it **changes which findings are suppressed**. It now reads `Narrow the suppression to line 40 alone by setting line_end = 40`, because `ash config lint --fix` should not quietly narrow suppressions across a repo while claiming to fix a mistake.

The fix itself is unchanged and still offered — a single line is usually what someone wants.

## Docs

The configuration guide only called `line_end` an "Optional ending line number", which is where the misreading starts. It now states the open-ended behaviour and shows the single-line form.

## Verification

- 7 new tests: 4 pin the matcher contract (pass on `main`), 3 cover the wording (fail on `main`).
- 129 passed across `tests/unit/config/`, `test_config_lint.py` and `test_suppression_matcher.py`.
- The new message deliberately keeps the `missing 'line_end'` phrase that the existing `test_lint_detects_missing_line_end` matches on, so no existing test needed editing.
- Docs freshness 7/7, schemas unchanged, ruff clean and formatted.

Fixes #347

---

## Update after review

The replacement wording was still not the predicate. `_line_range_matches` compares `finding_end >= suppression.line_start`, so a suppression at `line_start: 30` also matches a **multi-line finding that starts before line 30** and spans it — a finding at lines 10-50 is suppressed. `test_base_plugins_regression` already pins that behaviour.

Describing the range as "from line N to the end of the file" is closer than the original claim but still omits that, so someone auditing a suppression would not expect the earlier finding to be covered.

The message now says it matches any finding that *ends at or after* `line_start`, **including a multi-line finding that starts before it**, and a test asserts both halves of that sentence.


rafaelpereyra (commented):
Should this change require an update in the documentation too to clarify this?

rafaelpereyra (approved):

PR #465, opened 2026-08-26
feat(reporters): carry the project through every reporter in a workspace scan (phase 2b)

awsmadi:
## What this does

Workspace mode already scanned N projects and aggregated them (#462), but emitted no workspace-level *report*. The reason was one line in `github_ghas_reporter.py`: `run = sarif.runs[0]`. Against the N-run SARIF the aggregator writes, that emits the first project's findings and **drops every other project** -- no exception, no warning, just a smaller file a reviewer has no way to tell is incomplete. A security reporter that under-reports in silence is the worst failure this feature could ship, so the report step could not land until that was impossible.

Merging everything is not the fix. Merging is genuinely wrong for some formats: an SBOM for N independently versioned deliverables is N SBOMs, and a reporter that publishes a side effect would double-publish. So each reporter now **declares** what it does with a multi-project model, and `automated_security_helper/workspace/reporting.py` holds it to that declaration.

## The rulings

| Behaviour | Reporters | What you get |
| --- | --- | --- |
| `MERGED` | `sarif`, `html`, `markdown`, `text`, `csv`, `flat-json`, `yaml`, `junitxml`, `ocsf` | One artefact under `reports/`, carrying the project on every finding |
| `PER_PROJECT` | `github-ghas`, `gitlab-sast`, `cyclonedx`, `gitlab-cyclonedx`, `spdx`, `aws-security-hub`, `bedrock-summary-reporter`, `cloudwatch-logs`, `s3` | No workspace artefact. The files under `projects/<key>/reports/` are the answer |
| `WORKSPACE_SCOPED` | `unused-suppressions` | A workspace artefact covering workspace-level state only, not a merge |
| `UNSUPPORTED` | none | Refuses the run at exit 4 before anything is scanned |

The split is **9 merged / 9 per project / 1 workspace-scoped / 0 unsupported**.

Where the project appears depends on the format: a leading `workspace_project` column in `csv`, a field of the same name in `flat-json`, a per-project section in `html`/`markdown`/`text`, a `<project>/<scanner>` testsuite name in `junitxml`, and a `workspace_project:<key>` entry in `metadata.labels` for `ocsf`.

**The default is `PER_PROJECT`, deliberately fail-closed.** An undeclared reporter -- a third-party plugin, or a new built-in whose author did not consider workspaces -- is never handed a shape it has not said it understands, and loses nothing: its per-project artefacts are still produced and still correct. `MERGED` as a default would be fail-open, which is exactly the ghas bug for every external plugin. `UNSUPPORTED` as a default would fail an entire workspace run for any external plugin.

`reports/workspace-reports.json` accounts for every reporter considered, including the ones that deliberately produced nothing: its behaviour, the path of its workspace artefact or `null`, the per-project paths that replace it, which of those are missing on disk, and why a reporter was not considered at all (`disabled`, `dependencies-unsatisfied`, `not-in-requested-output-formats`). A missing report is never silent.

## Acceptance criteria

**23 -- every reporter has an asserted workspace behaviour.** Asserted twice over. `test_reporter_workspace_behaviour.py` pins each declaration against a transcribed copy of the RFC table, discovered by walking the plugin packages so a 20th reporter fails the count rather than inheriting a default nobody chose. `test_workspace_reporting.py` then pins that the driver *acts* on each behaviour, over all nineteen injected rather than a sample.

**No shipped reporter needs `UNSUPPORTED`** -- every one has a defensible merged or per-project answer. `cyclonedx` looked like a candidate, because a workspace-level call would `AttributeError` on the missing `cyclonedx` field, but the honest ruling is per project, and per project means it is never invoked at workspace level, so the crash is unreachable. **Read "0 unsupported" as a finding about the RFC, not as unfinished work:** the refusal mechanism is built and tested with a declared-unsupported test double, because an untested path is an assumed path.

**24 -- SARIF runs extractable per project.** One run per project, each with its own `originalUriBaseIds`, and `workspace.projects[i].sarif_run_index` selecting one. Asserted on the `sarif` reporter's own output -- not only the aggregator's, which #462 covered -- because the reporter dumps `model.sarif` and `SarifReport.merge_sarif_report` is a method on that very object that would collapse the runs. Re-asserted on disk by the end-to-end gate, which also checks the run roots are distinct.

## Breaking-ish: the empty-scan CSV header changes for single-directory scans

`CsvReporter` had two code paths that emitted **different headers for the same data**. A scan with findings emitted `FlatVulnerability`'s field names; a scan with none emitted a hand-written title-case list:

- **Before, no findings:** `ID,Title,Description,Severity,Scanner,Scanner Type,Rule ID,...`
- **Before, with findings:** `id,title,description,severity,scanner,scanner_type,rule_id,...`
- **After, both:** `id,title,description,severity,scanner,scanner_type,rule_id,...`

A consumer parsing by header name therefore worked against one form and not the other, depending on whether the scan happened to find anything. Both headers are now derived from `FlatVulnerability`'s field order and cannot disagree.

This was necessary for the workspace `workspace_project` column: a column a consumer cannot find in both forms is not a column. `test_report_empty_model_returns_header_only` pinned the inconsistency and is updated, with the reasoning in its docstring.

Anyone parsing empty ASH CSVs by title-case header name will need to switch to the field names the populated form has always used.

## Discovered pre-existing defect: the S3 object key is the constant `ash-report-None.json`

Found while checking the timestamp resolution behind the workspace collision. **Recommending this as its own decision -- it is not fixed here.**

`S3Reporter` builds its key as `f"{key_prefix}ash-report-{timestamp}.{ext}"` from `model.metadata.summary_stats.start`. That field is **not set when a reporter runs**: `ScanExecutionEngine.execute_phases` assigns it in a `finally` block at `execution_engine.py:547`, and `ReportPhase` runs at `:492`. So every reporter observes `None`, and every scan computes the identical key `ash-report-None.json`.

Verified by probing `ReportPhase._execute_phase` during a real scan rather than by reading line numbers:

```
AS SEEN BY THE REPORT PHASE: start = None        type = NoneType
AFTER the scan returns:      start = '2026-08-26T00:38:09+00:00'
```

Two consequences:

1. **All N projects in a workspace computed the same key and overwrote each other.** Not a race that concurrency makes likely -- a certainty. `PutObject` overwrites silently, so N-1 projects' reports vanished with no message. **This PR fixes this**, by putting the project in the key when one is present.
2. **Every run of a single-directory scan also overwrites the previous one.** Orthogonal to workspace mode, and **deliberately not fixed here.** Both behaviours are defensible and either choice breaks someone: a constant key loses every run but the last, while a varying key breaks anyone treating that object as a stable "latest report" pointer and accumulates objects indefinitely without a lifecycle policy. That is a product decision about a customer-facing AWS integration and belongs to whoever owns it. Moving the timestamp assignment earlier was also rejected -- `summary_stats.start` is read by more than this reporter, so that has a far wider blast radius than the key.

Single-directory keys are byte-identical to before. `test_the_start_timestamp_is_unset_when_a_reporter_runs` pins the premise so that if the timestamp ever becomes available, the assumption fails loudly instead of silently producing a different key.

## Defect introduced and fixed in this PR: the pre-flight was calling AWS before every scan

Recorded because it was mine, not pre-existing, and because it is the kind of thing that gets discovered by someone else's bill.

The `UNSUPPORTED` refusal is a pre-flight: it runs before any project is scanned, so an operator learns immediately and -- more importantly -- so the refusal cannot contradict a results file that has already recorded an exit code. My first version answered "is this reporter enabled?" by calling `validate_plugin_dependencies()`, which for `SecurityHubReporter` and `BedrockSummaryReporter` means **live calls to Security Hub and Bedrock**. That put paid API calls on the path before every workspace scan with the report phase, for a check that cannot change the outcome: a reporter whose dependencies are unsatisfied is already skipped by the report step itself.

It surfaced because two unit tests that only read declarations were reaching a real AWS account and logging `Security Hub is not enabled in this region`.

The pre-flight now checks config enablement and the `--output-format` filter only. Pinned by **counting the calls to zero**, with the companion assertion that the report step *does* still validate -- the negative assertion alone would pass if the validation had simply been deleted.

One narrow behaviour changes on purpose: a reporter that is enabled, declares `UNSUPPORTED`, and has unsatisfied dependencies now refuses the run rather than being quietly skipped later. That is the right direction -- the operator's configuration asks for something the format cannot do, and telling them so should not depend on whether an unrelated credential happens to be present.

## Note: per-project report titles

`metadata.project_name` is now set per project in workspace mode, to the project's display label, so each project's own `html`/`markdown`/`text` report names that project instead of the constant `"ASH"`. It was `"ASH"` for every project because the value is hardcoded in `AshAggregatedResults`'s `default_factory` and has never been derived from `AshConfig.project_name` in any mode.

**Single-directory scans are unaffected**: `metadata=None` is passed, the field keeps its `"ASH"` default, and the reports are byte-identical. Verified on a real single-directory scan, not inferred from the code.

The carrier is `ASHScanOrchestrator.metadata`, a declared parameter of that class since its introduction, documented as "Additional metadata for the scan", and read nowhere until now. It is needed because a project inside a workspace is scanned as a complete single-project run -- its own model's `workspace` is `None`, so a project does not know it is in a workspace -- and without something on the metadata its reports cannot say which project they describe. That mattered concretely rather than aesthetically: it is what made the four side-effecting reporters' per-project artefacts indistinguishable.

## Two RFC deviations, after reading the code

**`junitxml` uses `<project>/<scanner>`**, not the project alone. Every CI front end that renders JUnit XML groups by suite name, so the literal reading discards the per-scanner grouping single-directory mode has. Project leads, so one project's suites sort together, and the workspace artefact becomes a strict refinement of the per-project ones rather than a lossy reshape.

**`ocsf` uses `metadata.labels`**, not a `metadata` field. OCSF's `Metadata` sets `extra="forbid"`, and pydantic's `model_copy(update={...})` **silently drops** an undeclared key rather than raising -- so the obvious implementation of "project carried in metadata" would have shipped every finding unattributed with nothing to show it had failed. `labels` is the schema's own free-form slot and is SIEM-indexed. A test asserts no finding carries more than one project label, because one `Metadata` instance is shared across the report and mutating its `labels` in place would give every finding every project's label.

## Related commits and PRs

- **`b4c5963` is also PR #463**, cut against main because it fixes a live defect there: `tests/integration/cli/conftest.py` marked any test named `*workflow*`, `*lifecycle*` or `*end_to_end*` as slow, repo-wide, and `tests/conftest.py` then skips slow tests. Four unit tests are silently skipped on main today -- including `test_external_target_scan_gate.py`'s timeout guard, which has never run in CI since it landed. The commit is retained here so that this PR's own `test_the_merged_artefact_covers_both_projects_end_to_end` is not silently skipped in CI; the identical commit drops on rebase once #463 merges. **Not accidental duplication.**
- **`4851e5c` is also PR #464**, cut against main. `yaml.dump(model.model_dump(by_alias=True))` without `mode="json"` writes `!!python/object` tags, which `yaml.safe_load` refuses outright -- so `ash.yaml` could only be read by a loader that constructs arbitrary Python objects from the file. That is a security-relevant thing to require of anyone consuming a security tool's output. It affects **every document ASH has ever written**, empty ones included: a default `AshAggregatedResults` with no findings emits four such tags from the `AnyUrl` values in the SARIF tool metadata. Fixed in `spdx_reporter` too, same line. Retained here for the same reason as `b4c5963` -- the workspace `yaml` assertions depend on it -- and it drops on rebase once #464 merges.

## Stated limitations

- **Workspace-level reporter enablement comes from ASH's default config plus `--output-format`**, not from any project's config, because there is no workspace-level config until Phase 3. So a project that disables `html` still contributes to the workspace-level `html` report; its own `projects/<key>/reports/` respects its config as usual. The alternative -- withholding a merged artefact whenever any project disabled that reporter -- was rejected because it lets one project's configuration silently govern a workspace-level output.
- **`unused-suppressions` reports zero workspace-level suppressions**, because none can exist until Phase 3. It states its scope explicitly and names the per-project reports rather than being absent, because read as a merge its zero counts would mean "nothing unused anywhere" when what they mean is "no workspace-level suppression went unused".
- **The merged reporters need the whole model**, so the report step reads the unified file back as one `AshAggregatedResults`, giving up #462's streaming bound. Acceptable because it happens after every scan has finished -- the concurrency-scaled peak is unchanged and the two peaks do not add -- and because generating the reports from the file that was written makes it impossible for them to disagree with it.
- **A default run's manifest lists 15 reporters, not 19.** The four under `ash_aws_plugins` are registered only when an operator names that module in `ash_plugin_modules`. The manifest is exhaustive over what was registered, not over what exists on disk; the 19-reporter assertion covers the rest.

## Verification

- **Full suite on 3.13: 4207 passed, 114 skipped, 0 failed** -- identical under `-n auto` and `-n 0`. This is CI's scope (`pytest` with no path argument; `pytest.ini` sets `testpaths = tests`).
- **Full suite on 3.10: 4206 passed, 115 skipped, 0 failed.**
- On the narrower `tests/unit` scope: base is 4098 passed / 22 skipped, this branch is 4207 passed / 22 skipped. Zero new skips.
- **Self-scan: zero actionable findings on every scanner, none in ERROR.**
- `scripts/verify_docs_freshness.py` 7/7. `cz check` successful over all ten commits.
- **A real end-to-end gate**: `scripts/verify_workspace_reports.py` builds a two-project workspace, runs a real scan with the default phases, and asserts from the output tree that 9 merged artefacts exist with every project named, that 4 per-project rulings produced no workspace artefact and that their per-project files do exist, and that each project's own results carry a distinct identity.
- **32 mutations, all caught, across three harnesses each with a negative control** confirming the unmutated tests are green -- so "found nothing" is distinguishable from "checked nothing". 23 against the unit suite, 8 against the end-to-end gate, 1 against the prewarm/pre-flight ordering.

The mutation harnesses themselves are **not committed**, so that count is not reproducible from this branch. They mutate source files in place and are development scaffolding rather than product, so shipping them would add maintenance surface for little gain. The durable artefacts are the tests the mutations produced, which are committed and readable -- and the two findings below, which exist because a mutation caught them. Anyone wanting to re-derive the confidence can mutate the assertions directly; the point of recording the count is that no test in this PR was accepted without being watched failing, not that a reviewer can re-run a script to confirm it.
- **ruff 0.11.8** (the version pinned in `.pre-commit-config.yaml`, not the venv's newer one) over the diff's Python files: 8 errors and 5 format complaints, **every one verified present at the base ref** by extracting it to a temp tree and running the same commands. Zero new.

Two of the mutation findings are worth naming, because both were invisible from a green run:

- The end-to-end gate iterated `per_project_artifacts` asserting each path exists -- **vacuous on an empty list**. A mutation that routed per-project reporters down a different branch left the list empty while every other assertion still held, and the gate reported PASSED. It now asserts the pointer count and the project names before the paths.
- `test_every_project_appears_with_its_verdict` asserted the project name but **not the verdict**, despite its name. A mutation swapping `exceeds_threshold` for `finding_count` escaped it. The fixture is now asymmetric -- `web` has a finding but none actionable -- so the verdict cell can differ, and it is asserted.

One measurement artefact worth passing on: `ruff check` handed a non-Python path (the docs `.md` file) silently falls back to linting the whole tree and reported 4181 errors, where filtering to `.py` gives the real 8. Same family as `grep -cE "^FAILED"` matching nothing against pytest's ANSI colour: the tool answered a different question than the one asked, and did so confidently.

### DIFF
--- a/.github/workflows/ash-unified-ci.yml
+++ b/.github/workflows/ash-unified-ci.yml
@@ -266,6 +266,78 @@ jobs:
           if-no-files-found: warn
 
   # ---------------------------------------------------------------------------
+  # Multi-project attribution: assert a workspace scan keeps its projects apart.
+  #
+  # The sibling of external-target-scan, one dimension up. That job proves ASH
+  # can scan a directory that is not its working directory and asserts on what
+  # came back; this one proves that scanning N projects at once does not mix them.
+  # The four regressions it exists for all produce a complete, well-formed results
+  # file, so none of them can be caught by an exit status:
+  #
+  #   - a finding from project A reported against project B
+  #   - project A's suppression silencing the same rule at the same relative path
+  #     in project B, which is a silent false negative
+  #   - both projects judged against one threshold, so one verdict is wrong
+  #   - two projects sharing an output directory, the second overwriting the first
+  #
+  # The fixture puts a byte-identical file in all three projects, which is what
+  # makes attribution testable: with three different files, "attributed to A"
+  # would be satisfied by matching on the filename. One project is named 'src' on
+  # purpose -- container mode mounts the workspace root at /src, so that project
+  # is /src/src, the basename-collision shape from #361.
+  #
+  # Windows is in the matrix deliberately: this project's path, encoding and
+  # console bugs concentrate there, and workspace mode adds a whole path space.
+  # ---------------------------------------------------------------------------
+  multi-project-attribution:
+    name: "multi-project attribution (${{ matrix.os }})"
+    runs-on: ${{ matrix.os }}
+    # Keep in step with JOB_TIMEOUT_BUDGET_SECONDS in the script, which a unit
+    # test ties to this number. DEFAULT_SCAN_TIMEOUT_SECONDS must stay below it so
+    # the script's own diagnostics win the race against a job cancellation.
+    timeout-minutes: 25
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+      - uses: ./.github/actions/setup-ash
+        with:
+          python-version: "3.12"
+          install-ash: "true"
+      # bandit is the only producer this gate asserts on. Installing it here means
+      # an install flake fails as a failed install step with its own log, rather
+      # than surfacing as "the fixture produced zero findings".
+      - name: Install scanner dependencies
+        shell: bash
+        run: ash dependencies install
+      - name: Verify a workspace scan keeps its projects apart
+        shell: bash
+        # The gate builds its workspace with tempfile.mkdtemp(). Pointing the temp
+        # dir at the runner's scratch space makes the tree's location predictable
+        # enough to upload below. All three names because tempfile checks TMPDIR
+        # first and TEMP/TMP on Windows. This sits on the step rather than the job
+        # because the runner context does not exist in job-level env -- a workflow
+        # using it there is rejected outright, with no jobs and no annotation.
+        env:
+          TMPDIR: ${{ runner.temp }}
+          TEMP: ${{ runner.temp }}
+          TMP: ${{ runner.temp }}
+        run: python scripts/verify_multi_project_attribution.py --timeout 1200
+      - name: Collect the kept workspace and scan output
+        if: failure()
+        uses: actions/upload-artifact@v7
+        with:
+          name: multi-project-attribution-evidence-${{ matrix.os }}
+          path: ${{ runner.temp }}/ash-multi-project-*
+          retention-days: 7
+          if-no-files-found: warn
+
+  # ---------------------------------------------------------------------------
   # Install methods: method x OS x Python (matches ash-install-methods.yml)
   # ---------------------------------------------------------------------------
   install-validation:
--- a/automated_security_helper/base/reporter_plugin.py
+++ b/automated_security_helper/base/reporter_plugin.py
@@ -1,8 +1,44 @@
-"""Module containing the ReporterPlugin base class."""
+"""Module containing the ReporterPlugin base class.
+
+Workspace mode, and why a reporter has to declare what it does
+--------------------------------------------------------------
+A workspace scan produces one SARIF run per project, all in one
+``AshAggregatedResults``. That shape is new: every reporter in this repository
+was written when there was exactly one run, and several of them read
+``runs[0]`` and stop. ``github_ghas_reporter`` is the clearest case -- against an
+N-run model it emits the first project's findings and drops the rest, with no
+error, no warning, and a smaller output file that looks entirely plausible. A
+security reporter that under-reports in silence is the worst outcome workspace
+mode can have, and it is why no workspace-level report was emitted at all before
+this contract existed.
+
+The fix is not to make every reporter merge. Merging is genuinely wrong for
+some: an SBOM for N independently versioned deliverables is N SBOMs, and a
+reporter that publishes to a shared destination would double-publish. So instead
+each reporter *states* what it does with an N-project model, via
+:class:`ReporterWorkspaceBehaviour`, and
+``automated_security_helper.workspace.reporting`` holds it to that statement.
+
+The default is ``PER_PROJECT``, deliberately
+--------------------------------------------
+An undeclared reporter -- a third-party plugin, or a new built-in whose author
+did not think about workspaces -- gets ``PER_PROJECT``. That is the fail-closed
+choice and it is not a refusal: the reporter's per-project artefacts under
+``projects/<key>/reports/`` are still produced and still correct, because each
+project is scanned as a complete single-project run. All the default withholds
+is a workspace-level file the reporter has not said it can produce correctly,
+and the driver writes a manifest recording that withholding, so nothing is
+silent.
+
+``MERGED`` was rejected as the default because it is fail-open: it hands an
+unprepared reporter a shape it has never seen and trusts it. ``UNSUPPORTED`` was
+rejected because it would fail an entire workspace run for any external plugin.
+"""
 
 from abc import abstractmethod
 from datetime import datetime, timezone
-from typing import Annotated, Generic, TypeVar
+from enum import Enum
+from typing import Annotated, ClassVar, Generic, TypeVar
 from typing_extensions import Self
 
 from pydantic import Field, model_validator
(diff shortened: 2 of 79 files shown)


awsmadi:
## Please do not merge yet — three config fields are inert

This branch now also carries the Phase 3 workspace policy work (#472 was merged into it). Three fields in that work **parse, validate, appear in the generated JSON schema, and show up in the `--dry-run` plan, but never reach any scanner**:

- `workspace.suppressions`
- `workspace.ignore_paths`
- `workspace.additional_scanners`

An operator who writes a workspace-level suppression today would get no error, see it echoed back in the plan, and reasonably believe findings were suppressed. Nothing would be suppressed. A config field that validates and silently does nothing is worse than one that does not exist, and for suppressions specifically it is a false negative in a security scanner.

The severity ceiling (`max_severity_threshold`) is **not** affected — it is fully wired and does change verdicts, because the verdict is computed in `workspace/execution.py` from the plan rather than from the resolved config.

### Why they are not wired yet

The scanners read suppressions and ignore paths from the resolved `AshConfig`, and there is currently no way to hand the orchestrator one:

- `ASHScanOrchestrator.__init__` calls `resolve_config` itself and unconditionally reassigns its own `config` field, so a caller cannot supply a pre-resolved config.
- The only other channel, `config_overrides`, **fails open**: `apply_config_overrides` logs a warning and returns the *original* config when an override fails to parse or the merged result fails validation. Routing a severity policy through that means a typo scans with no policy while the operator believes a ceiling is enforced — the exact failure this feature exists to prevent.

Deliberately not worked around. The fix is to let the orchestrator accept a pre-resolved config, which is a change to the single-project path every non-workspace user depends on, so it is going up as its own PR against `main` where it can be reviewed on its own merits rather than buried in a diff about `.code-workspace` files.

### What unblocks this

1. The orchestrator change lands on `main`.
2. This branch rebases onto it and wires the three fields through.
3. The end-to-end workspace policy check is extended to assert a workspace-level suppression actually silences a finding in one project and not another.

Until then the three fields should be considered declared-but-unimplemented. Happy to be overruled, but please do it knowingly rather than by clicking merge on a green PR.

PR #493, opened 2026-08-28
feat(mcp): expose workspace mode over MCP with confined scan targets

awsmadi:
Workspace mode shipped in 3.7.0 and is CLI-only. `cli/mcp_server.py` registers 14 tools and mentions `workspace` zero times, so an agent driving ASH over MCP can scan one directory and nothing else. This adds two registered tools:

| tool | what it does |
| --- | --- |
| `resolve_ash_workspace` | Resolves a `.code-workspace` into a plan and returns it rendered. Scans nothing. |
| `run_ash_workspace_scan` | Scans every project under its own config and threshold, registers one scan per project, returns the per-project verdict. |

## The constraint that shaped the design

`interactions/run_ash_scan.py` dispatches the workspace branch and then calls `sys.exit` when the exit code is non-zero, and `_run_workspace_mode` exits on all three of its failure paths. `SystemExit` derives from `BaseException`, so it escapes the `except Exception` in `cli/mcp_tools.py` — a malformed workspace file would have taken down the MCP server process.

The workspace layer underneath is exception-based: `resolve_workspace` and `execute_workspace` raise and return, and neither calls `sys.exit`. So these tools call that layer directly and never enter `run_ash_scan`. Every failure arrives as an exception that becomes an error response, with `WorkspaceDefinitionError` mapping to 4, `ASHConfigValidationError` to 3, and anything else to 1, using the codes in `core/constants.py`.

There is a test asserting a malformed workspace file returns an error response and does **not** raise `SystemExit`. That assertion is the reason the design looks the way it does.

## Confinement

A workspace file names N directories, so without a check it would be an N-way arbitrary-directory-read primitive. Every resolved project directory goes through the existing `cli/mcp/scan_target.py::validate_scan_target`; a single refusal refuses the whole workspace and names the offending project key.

The workspace file's own path is deliberately **not** confined — it is a config input, the same category already left uncovered so that a profile config outside the allowed roots keeps working. Two tests pin that asymmetry. One allowlisted root covers all N projects by containment, so operators do not enumerate projects.

## Two changes outside the feature, each its own commit

**`build_project_scan_settings`** — `ProjectScanSettings` has 24 fields and was built inline in one place. A second caller rebuilding it by hand would silently miss any field added later, so the construction is now a shared builder. It owns both derived inputs, since `max_parallel_projects` and `project_timeout` come from a config read above the constructor and the phase list is assembled separately. A test asserts both callers receive the same object rather than comparing two constructions, because after the refactor an equality test compares the same code to itself.

**Canonical directory comparison in `ScanRegistry`** — the duplicate-scan check compared un-canonicalized path strings, so `/proj` and `/proj/` both registered and then raced on one output tree. Two tests failed against this before the fix. Canonicalization is applied to `get_scan_by_directory` and `list_scans` as well: fixing only the duplicate check would leave `register_scan` refusing a path while a lookup for the same string reported no such scan, and two code paths disagreeing about identity is worse than the original bug.

Separately, `validate_severity_threshold` rejected `ALL`, which is a legal value of `global_settings.severity_threshold`. A project declaring it could not register truthfully, and reporting it as `MEDIUM` would understate the strictest setting in the loosening direction. The validator now accepts it.

## Progress

`ctx.report_progress` has three fields and no project dimension. The project key travels in the message and progress is reported as completed projects over total projects. Per-project scanner fractions are not summed: each project's own denominator only grows and is capped below 1.0, so a sum would never reach completion.

## Not in scope

- Container mode. The workspace branch is gated on the run mode not being container; this path is local only.
- The source-delivery and profile tools that exist in `cli/mcp_tools.py` without a registration, and the session plumbing around them. Orthogonal subsystem.
- IDE auto-detect. There is no editor extension in this repository.

## Verification

Full suite **4591 passed, 114 skipped**, identical serially and under `-n auto` — worth stating separately because `addopts` carries `-n auto`, so the two are different runs and a parallel-only green is not a green.

Touched files also pass on a 3.10 interpreter. Documentation freshness is 8/8 — the check derives the MCP tool list from `mcp_server.py` and requires the README table to match, which is why the two rows are in this branch. Generated schemas are unchanged, and `ruff format --check` is unchanged on the two files it already failed on before this branch, with every remaining hunk outside this diff.

The 61 specifying tests were written before the implementation and committed red — 31 failed and 26 errored against a tree where the module did not exist yet.


automated_security_helper/cli/mcp/workspace.py
```diff
@@ -0,0 +1,749 @@
...
+    """
+    monitor: Optional["asyncio.Task[None]"] = None
+    if progress_reporter is not None and project_outputs:
+        monitor = asyncio.create_task(
+            monitor_workspace_progress(progress_reporter, dict(project_outputs))
+        )
+    try:
+        # Resolved from this module's globals at call time, so a test that
+        # replaces cli.mcp.workspace.execute_workspace is what runs.
+        return await asyncio.to_thread(execute_workspace, plan, settings)
```
nikmeiser, line 583: `offline=True` is a silent no-op over the workspace-MCP path (correctness + unexpected egress).

`cli/mcp/workspace.py` `_execute` calls `asyncio.to_thread(execute_workspace, …)` (`:557`) and never sets `ASH_OFFLINE`; the flag flows to `execution.py:579` → `orchestrator.py:99` (declared, never consumed). Scanners gate on `is_offline_mode()` (`core/constants.py:186`) reading `os.environ["ASH_OFFLINE"]` (`:188`). The CLI sets/pops it with try/finally (`run_ash_scan.py:680/:703` workspace, `:366/:509` local); the MCP path bypasses that. No test covers offline over the MCP tools. Fix: set `ASH_OFFLINE` around `execute_workspace` (try/finally) or honor `settings.offline` in the orchestrator.

awsmadi: Fixed in 3ade5e8e. Your trace was exact — the flag reached `ProjectScanSettings.offline`, reached the orchestrator's declared field, and died there, while `is_offline_mode()` reads only `os.environ["ASH_OFFLINE"]`.

`_execute` now sets it around the `execute_workspace` call.

**One deviation from the CLI's pattern, deliberately.** The CLI pops `ASH_OFFLINE` unconditionally when it's done, which is safe for a process that runs one scan and exits. This server is long-lived, and one started with `ASH_OFFLINE=YES` in its own environment — a reasonable choice for an air-gapped host — would have that deployment-level setting silently cleared by the first offline workspace scan. A request asking for the *safer* posture would have flipped the server to the less safe one, permanently. So the prior value is snapshotted and restored, and only removed when there was nothing there to begin with.

**A limitation I've stated in the docstring rather than half-fixed.** `ASH_OFFLINE` is process-global and this server can run sessions concurrently, so two overlapping workspace scans with different `offline` values share one variable and the first one's restore can land inside the second one's scan. That hazard is *newly reachable* here — it doesn't exist for the CLI, which is one process running one scan — so copying the CLI's pattern imports a race rather than inheriting one. Your second option, honouring `settings.offline` in the orchestrator, moves the same env-var write one layer deeper without removing it; closing it properly means having scanners consume the resolved setting instead of the environment, which is a change across every scanner. Serialising workspace scans behind a lock would make concurrent sessions wait minutes on each other, which is worse than the race it prevents.

**Six tests**, in `tests/unit/cli/mcp/test_workspace_offline_env.py`. The load-bearing one records the variable's value *during* execution, because that's the only moment a working fix and a no-op differ — reading `os.environ` after the call returns cannot fail, since a correct implementation restores it and a broken one never set it.

Three mutations, all killed, each by exactly one test: never setting it reproduces the no-op; always setting it regardless of `settings.offline` fails the negative control, which is what stops a "fix" that quietly cuts network access for every scan; popping unconditionally fails the ambient-value test. `settings.offline` still reaching `execute_workspace` is pinned separately, so if the orchestrator later honours it directly that doesn't have to be reinstated.

nikmeiser (approved):

PR #494, opened 2026-08-28
feat: distributed execute-and-collect, deployable AWS targets, and ASHScanStep

awsmadi:
## What this adds

Four things, in dependency order:

1. **Distributed execute-and-collect in ASH core.** `ash scan --shard-index/--shard-count` splits one scan across N executors, and a new `ash merge` recombines the results into a single report.
2. **Reference CI pipelines for distributed mode**, alongside the existing one-shot examples, using each platform's native fan-out (GitHub Actions matrix, GitLab `parallel:`, CodeBuild batch matrix, Jenkins `parallel`, CircleCI `parallelism`).
3. **Deployable AWS targets** under `deploy/`, in both CDK and Terraform: AgentCore Runtime, ECS Fargate, a Lambda one-shot CodeCommit pull-request gate, and a CodePipeline distributed executor.
4. **A jsii package** exporting `ASHScanStep` for `aws-cdk-lib.pipelines`, plus a generated buildspec for non-CDK users.

Also adds `--stateless-http/--no-stateless-http` and `--allowed-host` to `ash mcp`.

## Sharding is on the scanner axis, not the file axis

Worth calling out because it deviates from the obvious design. Partitioning by file path does not work here: only 6 of ~18 scanner plugins consult `scan_set` at all, and `orchestrator.source_scan_set` is write-only. A file-partitioned shard would silently run the full scan in every executor.

Instead, each shard is assigned a disjoint subset of *scanners*. A shard is applied as an **exclusion**, so it inherits `ScanPhase`'s already-tested SKIPPED/excluded bookkeeping rather than tripping its three completeness validators.

Partitioning is round-robin over a deduplicated, sorted, lower-cased scanner list, so input order cannot affect the split and every executor computes the same assignment with no coordination.

Consequence to be aware of: shard wall-clock is bounded by the slowest single scanner, so shards are uneven by nature. Sharding beyond the scanner count gains nothing.

## Per-shard exit codes must not be used as a gate

Each shard exits 0 even when it holds actionable findings — the verdict belongs to the collect step. This is by design, and it is measurable: in a 4-shard run over a fixture with 29 actionable findings, all four shards exited 0 and only `ash merge` failed. One shard owned two scanners, found nothing, and exited 0 while a sibling held 16 critical findings.

Gate on the merged report, never on a shard.

## Pre-existing defects fixed along the way

Found while building this, each independent of the feature:

| Defect | Effect |
|---|---|
| `ash mcp --host` never reached the MCP app | `--host 0.0.0.0` returned HTTP 421 to every proxied request |
| `mcp_list_scanners` returned an incomplete set | caused by memoization in `plugin_modules()` |
| `ash dependencies install` ran a `pip install` of an unclaimed PyPI name | executed by `cli/dependencies.py` |
| 7 of 11 output formats silently emitted nothing | `report_phase.py` compared `config.extension` against format *names* |
| `--min-severity` silently substituted a value | an unrecognized level fell back to rank 1 rather than erroring |

## The exit code could not distinguish "clean" from "never checked"

The most serious thing in this change set. `_compute_exit_code` derived its verdict purely from finding counts and never inspected scanner status, so a clean tree, a run where every scanner was MISSING, and a run where every scanner ERRORed all produced exit 0.

This is not theoretical. Deployed to Lambda, the CodeCommit gate commented "ASH scan passed" on a pull request containing two bandit HIGH findings — and with the approval gate enabled, called `update_pull_request_approval_state(APPROVE)` on code it never scanned. Root cause: Lambda requires a read-only root filesystem, and ASH's scanner toolchain writes to `$HOME` and various caches at scan time. Every uv-hosted scanner failed to acquire its lock; grype returned zero findings with no error at all.

Two guards were blind to the same condition from opposite directions: `verify_shard_coverage` checks six structural properties and never looks at scanner status, and the shard buildspec checked only that `ash_aggregated_results.json` exists — which a shard whose scanners all failed still writes.

`ash merge` inherits the same verdict function, so the distributed path had the identical exposure.

Fixed by adding an explicit way to require scanner completion. **The default preserves existing behavior**, because many legitimate environments lack cdk-nag, cfn-nag, grype, or syft, and flipping the default would turn currently-green runs red for reasons unrelated to security. The shipped deploy targets enable it, because a security gate should fail closed.

## No container image is published

ASH publishes no public image and this change does not add one. Every deployable target builds ASH into the adopter's own ECR repository via a CodeBuild project, which is also the bootstrap: the stack cannot come up until that build succeeds. This keeps the trust lifecycle of the scanning tool with the organization running it.

## Validation

Deployed and exercised in a sandbox AWS account, not just synthesized:

- **AgentCore** — real `InvokeAgentRuntime` round trips: `initialize`, `tools/list` returning 14 tools, `tools/call`, and a full `run_ash_scan` to `get_scan_progress` cycle completing in 88s with scan state surviving 18 separate invocations.
- **Fargate** — MCP server reaching "Application startup complete" and serving on `0.0.0.0:8000/mcp`.
- **CodeCommit gate** — a real pull request triggering the Lambda 18s after creation, with a comment posted back.
- **CodePipeline executor** — a 4-shard run with zero scanner overlap across all six shard pairs, union equal to the ten enabled scanners, and per-shard actionable counts summing exactly to the merged total.

Pre-deploy, a 3-shard run plus `ash merge` was shown identical to an unsharded scan of the same tree, and a merge missing a shard was shown to exit 1 naming the absent index.

## `McpStatelessHttp` defaults to true, and the reason is not the obvious one

AgentCore does **not** reject a stateful MCP server. Measured against a live stateful runtime, `initialize`, `tools/list` and `tools/call` all succeed, and controls confirm sessions are genuinely enforced (no session id returns 400; two differently shaped fabricated ids each return 404).

The real hazard is that AgentCore returns a *fresh* `Mcp-Session-Id` on every response while a stateful server honors only the one it issued at `initialize`. A client following AgentCore's documented guidance to reuse the returned id gets a 404 on its third call. Stateless is immune because it ignores session ids entirely.

## Known limitations

- Shard wall-clock is bounded by the slowest single scanner; a pathological scanner cannot be split.
- Three of five committed templates exceed CloudFormation's 51,200-byte `--template-body` inline limit and must be launched with `--template-url`. The README states which.
- The image is published behind a moving tag with no defined moment at which a running workload adopts a rebuild. Each build also pushes a version-qualified tag, and the stacks accept a parameter to pin it; the README documents the deploy-then-pin workflow.
- ASH requires `$HOME` and scanner caches to be writable at scan time. Any read-only environment, not only Lambda, needs those redirected.
- The jsii package builds and tests locally but publishes nowhere yet, so `ASHScanStep` is consumable only from a local build.
- `sse` transport is deprecated upstream. Fargate still permits it; the README steers to streamable-http.

## Reviewing this

It is large. If maintainers would rather review it in pieces, the natural split is ASH core (sharding, `ash merge`, the MCP flags, the exit-code fix), then `deploy/`, then `deploy/cdk-constructs/`, in that order — phase 1 is a hard dependency of the other two. Happy to restructure it that way.

CI additions: a drift gate that synthesizes every stack and byte-compares against the committed templates, `terraform fmt -check` and `validate` per module, cdk-nag over the synthesized assemblies, and a gate asserting no `aws` CLI invocation reaches the ASH image (which ships boto3 and no CLI).


awsmadi:
Merged `main` in to clear the conflict. Now `MERGEABLE`; full suite is **6244 passed, 116 skipped, 3 xfailed**.

The conflict markers badly understated the work. Of 208 files here and 45 on `main`, only 2 overlap and only one conflicted — and that was **inside a docstring**. The real incompatibility was in files that merged *cleanly*: 13 tests failed afterwards.

Those 13 all came from `main`'s ["test: raise package coverage from 85% to 93%" (#499)](https://github.com/awslabs/automated-security-helper/pull/499), which added three test files that don't exist on this branch's pre-merge head. They pass against `main` (107 passed) and 13 failed against this branch — so `main`'s tests are the contract here, except in one case where the test caught a real defect.

| failure | resolution |
| --- | --- |
| `'_Phase' object has no attribute '_shard_assignment'` | Fixed the **test stub**, not the engine. The engine reads it plainly on purpose, and the comment there explains why: a `getattr` fallback would let a refactor that stopped setting it degrade into stamping no shard metadata, i.e. a partial scan reporting itself as whole. The stub already models `_completed_scanners` for the same reason. |
| `TypeError: … unexpected keyword 'resolved_fail_on_incomplete_scanners'` | This branch added that kwarg to `_run_container_mode`; `main`'s three lambda doubles carry the old signature. Added explicitly rather than via `**kwargs`, so a *future* signature change still fails loudly. |
| `mcp_list_scanners` returned `[]` | This branch deliberately replaced `plugin_modules("scanner")` with `_loaded_scanner_classes`, because `plugin_modules` memoises into `_resolved_plugins` and never invalidates — a server was measured reporting 10 scanners while the registry held 13. Repointed `main`'s patches at the seam the tool actually uses. |

One of those tests was **not** repointed: its subject is that an installed distribution reaches the loader, so stubbing the function that calls the loader would delete the thing under test. It keeps the real implementation, and its expected argument list now covers the discovered package *plus* the vendored ones — which is what this branch passes, deliberately, since passing only the discovered set drops ferret, snyk and trivy the moment any `ash_*_plugins` distribution is installed.

**One product fix, because a `main` test caught a genuine defect.** `mcp_list_scanners` resolved `enabled` only inside `if raw_name:`, so a config declaring `enabled = False` was reported as **enabled** whenever its name was absent — telling a caller a scanner is participating when its own config switches it off. That resolution now runs regardless of the name. The same test exposed a gap on the instantiation-failure path, where `main` derived the config from `model_fields` and this branch reads `plugin.config` off a live instance: better when it works, but nothing when construction fails. `_declared_config_class` restores `main`'s fallback for that path only.

Two housekeeping notes:

- **`uv.lock` was regenerated.** git auto-merged it into a state describing neither side: `uv lock --check` failed on the merge result while passing on both `origin/main` and the pre-merge head. `uv lock` moves three packages (cryptography 50.0.1, gitpython 3.1.61, uv 0.12.9) — all versions the merged `pyproject.toml` already requires after `main`'s bumps.
- **`ruff format` reflowed ~56 lines I didn't edit.** `origin/main`'s own copies of those files also fail `ruff format --check`, so the repo doesn't enforce it and this mixes a small formatting sweep in. Kept rather than unpicked. `ruff check` is unchanged at 118 pre-existing findings.

One run showed a single failure in `test_scan_set_is_silent_by_default`. It passes in isolation, passes with its whole file in order, and two other full runs of this identical tree are green — order-dependent output capture under random ordering, not from this merge.

awsmadi:
Two things on this PR.

**The code-scanning alert on `AshAgentCore.template.json:906` is a false positive.** Line 906 is a closing brace; the finding anchors at the end of the Lambda resource block, so the subject is `ImageBootstrapStarter2B9D5AC0` starting at line 861. Its entire environment is one variable:

```json
"Environment": {
  "Variables": {
    "PROJECT_NAME": { "Ref": "ImageBuild31565E2D" }
  }
}
```

A CloudFormation `Ref` to a CodeBuild project name, resolved at deploy time. There is no hard-coded value there to be a secret, nothing to rotate, and nothing that exists before the stack does. The rule fires on the presence of environment variables, not on their contents.

I have not pushed a suppression for it. It is an informational alert on a check that is not required, and pushing here would restart ~250 check runs on a head that is otherwise complete — a bad trade for silencing one bot row. If it becomes noisy across other PRs, the right shape is a per-file suppression on this template rather than a rule-wide one.

**This branch has re-conflicted, for the third time.** `mergeable` is now `CONFLICTING` and `mergeStateStatus` is `DIRTY`, while the head is unchanged at `bd065b6df7c2` — so nothing pushed here caused it. #508 merging into `main` did. That is the third resolution this branch has needed from an unrelated merge landing, and it will need another after the next one.

That is a property of the branch's size rather than bad luck: 208 files means it intersects nearly everything. Worth deciding whether to keep re-resolving on each `main` merge or split it, because the current pattern spends a full conflict resolution and a full suite run per unrelated merge, and each one is a chance to resolve something the wrong way. I will re-resolve it, but flagging the trend rather than treating each instance as a one-off.

deploy/terraform/modules/codecommit-gate/files/ash_pr_gate.py
```diff
@@ -0,0 +1,387 @@
...
+        scan_exit, output_dir, log_tail = run_scan(source_dir, min_severity, fail_on_findings)
+        counts = read_severity_counts(output_dir)
+    except Exception as exc:  # noqa: BLE001 - any failure here is an error outcome
+        LOGGER.exception("scan failed")
+        log_tail = f"{log_tail}\n{exc}".strip()
+
+    # ASH's exit code is the verdict. Note the counts read above are not consulted:
+    # comparing them against min_severity here would be a second implementation of
+    # a judgment ASH already made, free to drift from it.
+    if scan_exit == EXIT_CLEAN:
```
nikmeiser, line 365: **Terraform gate false-greens on Lambda's read-only FS → approves unscanned code.** Blocking (safe-default for an auto-approve control). 

First: the disclosure here is excellent — `ADOPTION.md` reproduces this exact fail-open on two infrastructures and recommends against enabling the flag until Lambda can run scanners. This comment is only about making the code fail closed, not about the analysis.

`scan_exit == EXIT_CLEAN` → outcome "pass" treats exit 0 as a clean scan, but ASH also exits 0 when scanners are `ERROR/MISSING` (no `--fail-on-incomplete-scanners` in the argv at `:194-208`). That "pass" then drives `update_pull_request_approval_state(..., "APPROVE")` at `:370-375` when `ASH_MANAGE_APPROVAL_STATE` is on — so an operator who arms the approval path on a Lambda where the scanners can't run gets auto-approval of unscanned code, which is precisely the incident `ADOPTION.md` measured ("arm B: merge exit 0, actionable=0").

A gate that can APPROVE must fail closed here. Two changes:
1. Add `--fail-on-incomplete-scanners` to the argv (`:208`) so exit 0 genuinely means every requested scanner ran — or, equivalently, gate the "pass" outcome on the per-scanner table showing no `ERROR/MISSING`. Right now the doc note is the only thing standing between an adopter and a false green.
2. Fix the root cause so the flag is actually usable: the scanners fail on Lambda's read-only FS because their caches have nowhere to write. The CDK gate already solves this (deploy/cdk/lib/ash-container-scripts.ts:515 _scan_env()` redirecting `HOME/XDG_CACHE_HOME/UV_CACHE_DIR/GRYPE_DB_CACHE_DIR` to `/tmp`, applied via `env=` at `:618`); the Terraform gate's `_run` (`:98`) passes no `env=`, and `main.tf:452-465` sets only `ASH_CONFIG_DIR`. Porting `_scan_env` here would let the scanners run, at which point enabling the flag keeps it that way.

Minimum to unblock is (1) — it removes the dangerous outcome regardless of environment. (2) is what makes the gate useful rather than just safe. Happy to see this split into a follow-up so the distributed core / `ash merge` (which are strong) can land separately.

deploy/cdk/lib/ash-container-scripts.ts
```diff
@@ -0,0 +1,661 @@
...
+    else:
+        tail = ((scan.stderr or scan.stdout or "").strip())[-2000:]
+        parts += ["ASH produced no markdown report. Tail of its output:", "", "\`\`\`", tail, "\`\`\`"]
+    parts += ["", "_Scanned commit \`%s\` against \`%s\`._" % (source_commit[:12], destination_commit[:12])]
+    body = "\\n".join(parts)
+
+    _post(codecommit, pull_request_id, repository_name, destination_commit, source_commit, body)
+
+    if os.environ.get("ASH_APPROVAL_GATE", "false") == "true" and revision_id:
+        state = "APPROVE" if verdict == "passed" else "REVOKE"
```
nikmeiser, line 634: **CDK REVOKEs on error (`ash-container-scripts.ts:634`), Terraform leaves approval unchanged**
(`ash_pr_gate.py:368-382`, now with a rationale comment but no reconciliation). Still no Terraform-gate
unit test (no Python test targets `ash_pr_gate.py`) and no CDK↔Terraform parity test.

Concrete scenario where it bites: a PR is APPROVED (by the gate on an earlier clean commit, or by a human), then a new commit introduces findings and the gate re-runs.
- CDK: findings → REVOKE → the stale approval is withdrawn; the PR is blocked. ✅
- Terraform: findings → leave unchanged → the old APPROVE stays in place on code that now has findings. The gate won't pull approval for newly-bad code.

deploy/terraform/modules/codecommit-gate/files/ash_pr_gate.py
awsmadi: Fixed in 40478712, and you were right that the doc note was the only thing standing between an adopter and a false green.

**Minimum-to-unblock, done:** `--fail-on-incomplete-scanners` is in the scan argv. I hardcoded it rather than exposing an environment variable — a gate whose fail-closed behaviour can be switched off by configuration has a safety property that varies by deployment, and the failure is silent when it's set wrong. There's a test asserting the flag survives alongside `ASH_SCAN_EXTRA_ARGS`, since that's the one knob that could smuggle in the negation.

Worth noting the flag's own help text in `cli/scan.py` describes exactly your finding — *"a run where nothing ran exits 0, the same code as a clean scan"* — and that it takes precedence over `--fail-on-findings` because "a partial scan's findings are real but its clean bill of health is not." So the mechanism was already there and this gate just wasn't using it.

**(2) taken as the follow-up you offered.** Porting `_scan_env` is what makes the gate useful rather than merely safe, and it's a bigger change than this one. This removes the dangerous outcome regardless of environment, which was your stated minimum.

**Tests, since there were none.** Eight in `tests/unit/deploy/test_codecommit_gate_fails_closed.py` — the first to target this file. Loaded by path, since it's packaged into a Lambda rather than importable, which is also what makes testing it possible at all. Three mutations, all killed, each by a different set: dropping the flag reproduces this defect, restoring leave-untouched reproduces the other one, and revoking on pass as well fails the clean-approval case — that last is the control that stops an over-correction reading as a fix.

One thing I'd flag from writing them. The handler builds its client inline with `boto3.client("codecommit")`, so patching a module attribute with `raising=False` silently did nothing and let the real factory run. Botocore then found my own `~/.aws/credentials` and failed on a missing region — a unit test one environment variable away from calling a live AWS API. The factory is patched now, with an assertion that no other service is requested.

deploy/cdk/lib/ash-container-scripts.ts
awsmadi: Fixed in 40478712. Your scenario is the one that decided it: an approval granted on an earlier clean commit surviving a later commit that introduced findings, so the gate declines to withdraw the very approval it granted, on code it has just judged unclean.

The old rationale — a transient infrastructure failure shouldn't look like a security judgment — is real, but it points the other way once approvals can already exist. A stale `APPROVE` is invisible to a reviewer reading the PR; an over-eager `REVOKE` is visible and self-correcting, because the next clean run re-approves. Asymmetric costs, so it revokes.

`REVOKE` now covers `findings` and `error` alike. `error` includes the incomplete-scanner exit added in the other thread, which is exactly the case where an approval must not survive — nothing was scanned, so nothing supports it.

**On the two gaps you named.** There were no tests on `ash_pr_gate.py`; there are now eight, including a parametrized table over exit codes 0/1/2/70 asserting the resulting approval state, plus a separate test for "a non-pass never leaves the state untouched" — that one is split out deliberately, because "no call was made" and "the wrong call was made" are different failures and only the first describes what you found.

The CDK↔Terraform parity test is **structural, not behavioural, and says so in its docstring.** The CDK handler is a Python string inside a TypeScript template literal that only executes inside a Lambda image, so I can't drive both from one test. What it checks is that neither file carries the shape you flagged — both have a `REVOKE` path, and the `leaving approval state unchanged` branch is gone. A text assertion can prove the branch exists, not that it behaves; recording it as the weaker check it is rather than claiming parity is proven.

awsmadi:
Pushed four commits: the `main` fold that clears the conflict, and the two review findings on the cdk-nag suppression matcher.

## Conflict resolution

`main` merged in. Two conflicts, both in dependency metadata.

**`pyproject.toml`**, the `[dependency-groups] dev` list. `main` lowered every floor from patch to major.minor and keeps patch precision only where an advisory forces it, with a trailing comment saying so. This branch's side was the pre-rewrite list plus one deliberate addition, `moto[server]`, for the subprocess tests under `tests/unit/deploy`. Took `main`'s list verbatim and re-added `moto[server]` at the end, written `>=5.2` rather than the `>=5.2.3` this branch had: nothing needs 5.2.3 specifically, `ThreadedMotoServer` and the `[server]` extra both long predate it, so under the new convention a patch floor here would imply an advisory that does not exist and Dependabot would treat moto's patch bumps differently from every other dev dependency. Flagging it since it is the one specifier I changed rather than carried across.

**`uv.lock`**. Rather than hand-merge a resolver artifact, took `main`'s refreshed lock whole and re-ran `uv lock`, which added moto and its transitive closure back on top. The result differs from `main`'s lock by exactly those additions plus one resolution-marker line that splits at 3.11 because `cfn-lint` pulls `networkx` at two versions — byte-identical to the split this branch's own lock already carried. `ruff` and `types-boto3` land on `main`'s pins rather than this branch's newer ones, which is the point of preferring `main`'s lock. `uv lock --check` is clean.

One piece of fallout, caught by a guard on this branch rather than by me. `_CDK_EXTRA_FALLBACK_REQUIREMENTS` in `cdk_nag_scanner.py` is a hand-maintained copy of the `cdk` extra, reached when the installed distribution's metadata cannot be read, and `test_fallback_matches_installed_metadata` compares the two. `main` lowered that extra's floors; the copy still had the patch-level ones, so the test failed on the merge — doing its job on a bound change that arrived from the other side of a merge rather than from an edit here. The constant does not exist on `main`, so there was no `main`-side version to prefer during resolution and the file merged cleanly; the mismatch only surfaced when the test ran. Fixed by moving the copy, not by relaxing the assertion. The three requirements still resolve to aws-cdk-lib 2.267.0, cdk-nag 3.0.2 and constructs 10.8.1, so this changes which versions are acceptable, not which ones anyone gets.

## Finding 1: the fail-closed warning named the wrong thing

Confirmed, and worse than reported. The regex-form warning in `_scope_covers_qualifier` fires per `applies_to` **element** and then continues the loop, so for a scope written

```yaml
applies_to:
  - regex: "/^Resource::/"
  - "Resource::*"
```

it logged `Ignoring the in-template cdk-nag suppression for '...' ... The finding is reported rather than silenced on a guess` and then returned `True` on the next element — suppression honored, finding silenced, log saying the opposite. Measured: the same text is emitted for a call returning `True` and for a call returning `False`, so the message carried no information about the outcome it was asserting.

The sibling non-list warning made a smaller version of the same claim. It does skip its whole entry, but `The finding is reported.` is still not its call to make: `_template_suppression_reason` keeps walking `rules_to_suppress`, so a later entry covering the rule suppresses the finding anyway. Measured with a scalar `applies_to` followed by a bare-id entry — warning emitted, finding suppressed. Fixed alongside, since it is one defect seen twice.

Both now name what was skipped and its scope (one element, or one entry) and stop asserting a final outcome neither of them can see. The regex one also says what to do about it: list the qualifier verbatim alongside the regex, since plain strings are compared exactly. No control flow change — failing closed on a JavaScript pattern this scanner will not reinterpret under Python's `re` is correct and is untouched. Only the text was wrong, and a message naming the wrong cause sends someone debugging a dropped suppression off to rewrite a regex that had nothing to do with it.

## Finding 4: the star-prefix mutation

Reproduced before fixing. Substituting

```python
if qualifier.startswith(member.rstrip("*")):
```

for `member == qualifier` and running

```
uv run pytest tests/unit/utils/test_cdk_nag_wrapper_behavior.py \
    tests/unit/plugin_modules/ash_builtin/test_cdk_nag_scanner_behavior.py
```

gives **119 passed**. The gap is not missing negatives — that section has several — it is that they all happen to hold their two strings in the prefix order a star-prefix rule answers `False` for anyway. The declared scope in them is a long bucket ARN and the reported qualifier is the short `Resource::*`, so stripping the star off the ARN yields a string `Resource::*` does not begin with, and every one of them still passes.

The direction that matters is the other one, and it is the common one on real templates. `Resource::*` is a finding *string* — cdk-nag's text for a permission on every resource — not a glob over finding strings. Under a prefix rule an author who reviewed and accepted that single wildcard finding would silently also suppress every specific-ARN finding on the same resource, each one carrying the wildcard's reason as its recorded justification. A suppression wider than it was declared to be hides real findings and misattributes why.

`test_a_scope_member_matches_the_qualifier_exactly_not_by_prefix` adds three parametrized cases, values from the vocabulary `AwsSolutions-IAM5` really emits: a declared `Resource::*` against a specific asset-bucket ARN, a declared `Action::s3:GetObject` against a reported `Action::s3:GetObjectVersion`, and that pair reversed. Both directions proven able to fail:

| | mutation above | mirror image `member.startswith(qualifier.rstrip("*"))` | correct code |
|---|---|---|---|
| `declared-wildcard-vs-a-specific-arn` | FAIL | FAIL | pass |
| `declared-scope-is-a-prefix-of-the-qualifier` | FAIL | pass | pass |
| `qualifier-is-a-prefix-of-the-declared-scope` | pass | FAIL | pass |

So no case is one that cannot fail, and the equality is pinned however a prefix rule gets spelled rather than against one spelling of it. 122 passed with the mutation reverted.

## Verification

Full suite, `uv sync --extra cdk` then `uv run pytest`, matching CI: **6536 passed, 0 failed, 119 skipped, 3 xfailed**, exit 0. Python coverage **93.16%** measured on its own; no TypeScript file is touched, so that gate is unaffected.

The 119 skips are all pre-existing and none are in this area: 89 behind `--run-integration`, 8 declared redundant with `TestFailClosedIsNecessary`, 5 behind `--run-slow`, 2 Windows-only on a Linux runner, and the remainder single tests carrying their own reasons in the tree already. The `--run-integration` set matters most here because it is the only path that runs the changed wrapper against real cdk-nag, so it was run separately the way CI does: `tests/integration/scanners/test_cdk_nag_real_pack.py --run-integration` with `ASH_REQUIRE_CDK_EXTRA=1`, **3 passed** against cdk-nag 3.0.2 / aws-cdk-lib 2.267.0.

`ruff format --check` clean on all three touched files under the lock's ruff, which is worth stating because preferring `main`'s lock moved ruff back a couple of patches and the last commit before the merge was a reformat. `ruff check` adds no finding: 22 before the change and 22 after on the wrapper, all pre-existing and outside the edited lines.

One honest loose end, reported rather than papered over. The first full-suite run also failed `tests/unit/utils/test_get_scan_set_coverage.py::test_scan_set_is_silent_by_default`, on two INFO records from `get_scan_set.py` reaching stdout. It passed on rerun with identical code, passes in isolation, and passes with the whole `tests/unit/utils/` directory run serially in one process, so it is order-dependent logger-handler contamination from somewhere outside that directory rather than anything reachable from this diff. `pytest-randomly` is not installed, so the varying factor is xdist `--dist load` deciding which tests share a worker; adding three tests changes that packing. I could not localize the contaminating test and did not chase it further here — it looks like a latent flake worth its own issue, not something this branch introduced.

rafaelpereyra (approved):

PR #500, opened 2026-08-31
ci: run the TypeScript suites, with the denominator pinned

awsmadi:
## The gap

`jest` appears nowhere under `.github/`, so neither TypeScript package under `deploy/` is tested in CI. Both have real suites and both measure above 90% statement coverage today — but nothing enforces either number, so a regression lands unobserved.

Measured with each package's own `npm test` plus `--coverage`:

| Package | Statements | Coverage | Lines | Tests |
|---|---|---|---|---|
| `deploy/cdk` | 383 / 405 | **94.56%** | 94.51% | 352 passed, 13 suites |
| `deploy/cdk-constructs` | 221 / 227 | **97.35%** | 97.33% | 116 passed, 3 suites |

Two separate packages with their own `package.json`, `jest.config.js`, and `node_modules`. There is no combined runner, so each is gated independently at 90%.

## Why a denominator check runs alongside the threshold

A percentage threshold on its own is satisfiable by measuring *less*, and that is not hypothetical in this repository.

Both jest configs set `roots: ['<rootDir>/test']`, which confines jest's crawler to the test directory. A source file that no test imports is then omitted from the report **entirely** rather than reported at 0%. `deploy/cdk` declared no `collectCoverageFrom` at all, so its denominator was 377 rather than 405, with 28 statements outside it. Nothing errored in either case.

`deploy/cdk-constructs` went through three states, and they are worth spelling out because two of them read as the same percentage:

| State | Covered / total | Coverage | Scope |
|---|---|---|---|
| Before the fix | 184 / 189 | 97.35% | narrow — `buildspec-cli.ts` absent from the report entirely |
| Scope widened, no new tests | 184 / 223 | **82.51%** | wide — the honest figure, and the real gap |
| Current (this PR's numbers) | 221 / 227 | **97.35%** | wide — after tests were added for that file |

184/189 is 97.354% and 221/227 is 97.357%, so the inflated reading and the corrected one round to the same two decimal places. That coincidence is exactly what makes this trap hard to spot from a percentage alone, which is the argument for pinning the denominator rather than only the percentage. The denominator moved 223 → 227 because making the file testable added four statements.

The diagnostic detail in the middle row: when the scope was widened, the **covered** count did not change at all — 184 before and after. Only the denominator did. So narrowing `roots` raises the percentage without touching a single test, while reading as a config tidy-up in review.

`.github/scripts/assert-coverage-scope.mjs` pins the measured scope — file count, statement denominator, and the presence of specific files by name. Shrinking what is measured now fails loudly instead of raising the number. The floors are measured values rather than estimates: adding source raises them legitimately, removing measured code lowers them and fails, which is the case worth catching and should require an explicit edit to the workflow.

The same shape has bitten this repository in Python: `automated_security_helper/tools/` has no `__init__.py`, so coverage.py's `source=` scan never descends into it and `install_dependencies.py` is absent from every report rather than reported uncovered.

## Proven able to fail

Controls run locally, all verified:

- The scope script fails on fewer files than required, a smaller denominator, an absent required file, a missing report, and a malformed argument — and passes on both real reports.
- jest's threshold rejects `statements>=99` against an actual 97.35%, and accepts the configured 90/90.

One silent-pass path was found and closed while building it: a non-numeric `--min-files` became `NaN`, and `length < NaN` is `false`, so an argument that lost its value would have skipped the check while still exiting 0. That is the exact failure the script exists to prevent, so it is now a hard error.

## Two deliberate choices, stated so they are not read as oversights

**Path filters** restrict the job to `deploy/cdk/**` and `deploy/cdk-constructs/**`, because `deploy/` does not exist on `main` yet — it arrives with #494. Letting the presence check skip quietly when the directory is absent was rejected: a job reporting success because it found nothing is indistinguishable from one that tested something. So the two behaviors are split — the filter decides *whether* the job runs, and the presence check makes a run impossible to pass without a package.

**This workflow and its script are not in their own path filter**, so the change adding them does not trigger a job against a tree with no `deploy/` and fail its own gate on arrival. Once `deploy/` is on `main`, adding those two paths back is worth doing so edits to the gate are themselves tested. Noted in a comment in the workflow.

## Known limitation

`bin/ash.ts` reads 0% and is required by name in the scope check anyway. It *is* executed — `test/ash-nag-gate.test.ts:156` runs it through ts-node in a **subprocess**, which istanbul cannot instrument from the parent. Genuinely exercised, unavoidably uncounted. `deploy/cdk` clears 90% without it, and naming it in the scope check is what stops it silently leaving the denominator.

## Sequencing

This targets `main` and is independent of #494. The job is inert until `deploy/` lands, at which point it activates on the next change to those paths. It was verified locally against #494's tree, which is where the numbers above come from.


nikmeiser (dismissed):
Approved with nits - 

- In addition to the caveats already documented there is one additional subtlety - a new file imported-but-thinly-covered still dilutes the global average and can pass
- not enforceable as a required check while path-filtered to `deploy/cdk/**` +
`deploy/cdk-constructs/**`

awsmadi:
Thanks — both land, and the second one is sharper than a nit. Taking them separately.

**On dilution.** Correct, and worth stating precisely so the boundary of what this PR buys is clear. `assert-coverage-scope.mjs` pins the denominator so it cannot shrink, and `assert-coverage-completeness.mjs` takes the census from `git ls-files` so a file no harness imports cannot go missing from the report — but neither is a threshold. `assert-coverage-completeness.mjs` says so in its own header: "This script has no thresholds." So a new file that *is* measured and thinly covered contributes its statements to the denominator and its low count to the numerator, and the global average absorbs it. Whether that passes depends entirely on how much well-covered code it is averaged against, which is not a property anyone reviewing the new file can see.

What these scripts close is the case where measuring *less* raises the number. What they don't close is measuring more, badly. A per-file floor is the fix, and it's a policy question this PR shouldn't answer by itself — floor at what percentage, applied to new files only or all files, and what happens to files that are legitimately thin. Filing rather than folding in.

**On enforceability — this is the one I'd act on first.** It's structural, not a caveat. The `pull_request` trigger is filtered to `deploy/cdk/**` and `deploy/cdk-constructs/**` (lines 104-109), so a PR touching neither produces **zero** runs of this workflow, not passing ones. A required check with no run sits pending indefinitely, so marking it required as written would stall every PR that doesn't touch those two directories — and stall the merge queue with them. The failure looks like an unrelated hang rather than a misconfigured gate, which is the part that makes it expensive to diagnose.

The usual way out is a job that always runs and owns the required check name, short-circuiting to success when the filtered paths are untouched and otherwise doing the real work — path filtering moves from the workflow trigger to a step-level condition. That's a restructure of the trigger, so also a follow-up rather than something to push onto an approved PR.

Neither changes the value of what's here: before this, `jest` appeared nowhere under `.github/`, so 468 tests across two packages ran in CI not at all. Advisory-but-honest beats absent. I'd just avoid adding it to the required list until the trigger is restructured.

awsmadi:
Correcting myself: I reported zero failing checks here, and that was wrong. There is one, and it was invisible to how I looked.

This head carries **152 check runs**, and the failing one sits past the first page. A single `per_page=100` read — which is what `gh pr view --json statusCheckRollup` gives you — returns only benign rows, so a conclusion tally over it says "clean". Paginating finds:

```
failure   unit-test (ubuntu-latest, py3.13)   id=100443175734
```

`mergeStateStatus` is the cheap tell I should have been reading: it's **UNSTABLE**, not BLOCKED. That distinction is precisely "a non-required check is unhappy" versus "a required one is", and it costs one field. A page-one tally structurally cannot produce it.

**What it is.** `tests/unit/workspace/test_execution.py::TestFailureHandling::test_a_project_timeout_fails_that_project_only`:

```
AssertionError: assert <ProjectRunStatus.FAILED> is <ProjectRunStatus.COMPLETED>
ERROR ash:execution.py:1052 Project 'quick' exceeded its 0.2s budget after 0.2s
ERROR ash:execution.py:1052 Project 'slow'  exceeded its 0.2s budget after 0.2s
```

The test runs `slow` (sleeps 1.5s) and `quick` (no sleep) at `project_timeout=0.2`, and asserts the timeout fails only `slow`. On this runner **both** exceeded the budget — including the one with nothing to do. So 0.2s is smaller than the baseline cost of starting a project there: thread startup plus orchestrator construction alone. Nothing to do with which project is slow.

I ruled out the other candidate mechanism rather than assuming: `FakeOrchestrator.behaviour` is a class-level dict, so a leaked entry would produce the same symptom, but it's reset in the fixture at lines 119 and 122 and no test configures a sleep on `quick`.

**It isn't this PR's.** This PR adds four files, all CI-side — two `.mjs` scripts, an exclusions manifest, and a workflow. No Python. The control is #514 and #493: both merged the same `origin/main` commit into the same base, and both have complete check-run sets with nothing non-benign, so the same test on the same code passed there. This is a pre-existing wall-clock flake that this run happened to land on.

**Where the fix goes.** Its own branch off `main`, not here. Worth noting the neighbouring test — `test_the_timeout_bounds_wall_clock_when_the_bound_is_smaller`, 25 lines down — already solved this class and says so in its docstring: its wedge "releases on an event rather than sleeping, so the test asserts the workspace returned while the worker was still running ... without depending on a sleep duration." This test still races a wall clock. The repair is to widen the ratio rather than tighten it: a budget comfortably above baseline startup overhead, with the slow project's sleep well above the budget. The property under test survives; only the timing coupling goes.

So on the question of merging over it — my read is that it's mergeable, because the red cell is neither required nor caused by this change, and it will keep firing on unrelated PRs until the test is repaired. But it should be repaired, and I'd rather that be tracked than absorbed.

awsmadi:
Correction to my previous comment: this PR touches **five** files, not four. I wrote "four CI-side files and no Python", and the fifth is `uv.lock` (+121 −121) — which I regenerated myself during the `main` merge and described in that merge's commit message, so I had no excuse for the miscount.

The gap that created is real and worth naming rather than waving off. `uv.lock` is not Python source, but it decides which package versions install, and a timing-sensitive test's startup cost is exactly the kind of thing that moves when the import graph changes. "No Python, therefore not the cause" did not cover it.

**It closes, and more firmly than a line count could show.** The lockfile is byte-identical across all three PRs and `main`:

```
#500         687029579a2f46d113e337f30037f82beacb90a4
#514         687029579a2f46d113e337f30037f82beacb90a4
#493         687029579a2f46d113e337f30037f82beacb90a4
origin/main  687029579a2f46d113e337f30037f82beacb90a4
```

One distinct git object across four refs. So there is no dependency-environment difference between this PR and the two that pass the same test — nor between any of them and `main`. The control holds that variable constant rather than my having excluded it by inspection.

**One instrument note, because the tempting shortcut here is wrong.** Matching `+121 −121` on all three is not what establishes this. A PR's file diff is computed against *its own merge base*, and these three branches were cut at different times, so equal tallies across them would be coincidence — two lockfiles can churn the same number of lines and resolve different versions. The blob id is what settles it. It happens to agree here; the counts could not have told you that.

The flake diagnosis is unchanged: `project_timeout=0.2` with both `slow` and `quick` exceeding the budget, `quick` configured to sleep not at all, is a wall-clock problem in the test rather than anything about this change or about Python 3.13.

awsmadi:
CI fixed. Merged `main` (3 commits, needed anyway — the ruleset sets `strict_required_status_checks_policy`, so a branch can't merge while behind) and merged `fix/workspace-timeout-test-wall-clock`, which carries the fix for the failing cell.

**Heads up: this push dismissed the approval.** The ruleset sets `dismiss_stale_reviews_on_push` with `required_approving_review_count: 1`. That cost was unavoidable — the strict policy meant this branch had to merge `main` before landing regardless.

The failure was `unit-test (ubuntu-latest, py3.13)` on `test_a_project_timeout_fails_that_project_only`, which this branch doesn't touch. A 0.2s project-deadline budget smaller than the fixed cost of starting a project on a loaded runner, so the project configured to do *no work* also overran and the "fails only the offending project" assertion broke. Fix is in #521; merging that branch here makes this branch deterministic rather than waiting for it to land. Same commit either way, so it converges with no conflict.

Full suite: **5813 passed, 114 skipped, 3 xfailed**, coverage 93.31%. The previously-failing class is 12 passed.

`uv lock --check` still fails here, and for once it isn't merge drift — it's exactly the two-entry staleness `main` already carries (`boto3` and `types-boto3`, both `>=1.43.80` against a pyproject asking `>=1.43.81`), measured as the complete `uv lock` diff rather than grepped. Inherited, not created here, so it's left for its own fix; the incidental relock `uv run` performs was reverted.

---

**On why that check wasn't enforced** — it's structural, and it's the same problem class as the nit above about this workflow's path filter.

Every job in `ash-unified-ci.yml` names itself from its matrix, so the workflow contributes 84 distinct check names and not one is stable. On this head: **152 check runs, 2 required**. `required_status_checks` holds literal context strings with no wildcard, so requiring `unit-test` means enumerating all 18 of its generated names — and any matrix edit then either orphans a required context that no longer runs (pending forever, wedging the merge queue) or adds a leg nothing requires. `lint` is the only stable name in the file and isn't required either.

#522 adds a `required-checks` job with a matrix-independent name that can be required once and stay correct. Two fail-open shapes are closed in it: `if: always()`, because a skipped required check counts as a pass and a failed dependency would otherwise skip the gate; and an asserted upstream count, so a job added to the workflow can't escape the gate silently. Adding it to the ruleset is an admin action and deliberately left out of that PR.

nikmeiser (approved):
Non-blocking — this gate is a clear improvement and the reasoning is excellent (the `min_statements/min_files/--require` denominator pin closes "measure less," and the git-census in `assert-coverage-completeness.mjs` closes "lost a file"). One residual it doesn't cover: `{"global":{...}}` here (and at `:150`) is a global floor, so a new source file that some test imports but barely exercises dilutes into the aggregate and passes as long as the global stays ≥90%. The denominator pin defends against files omitted or dropped; it doesn't require any individual file to hit a bar. A large thinly-covered file would move the aggregate and get caught — it's small additions, or existing slack above 90%, that slip through.
Your header comment enumerates "measure less" and "lost a file" as the two gaming vectors; this is a quiet third one. If you want to close it, a jest per-glob `coverageThreshold` (e.g. `"./src/**": {statements: 85}`) or a per-file floor alongside the global would do it. Fine as a follow-up — not a blocker for landing the gate.

awsmadi:
Thanks for the re-approval, and your third gaming vector is correct — I'd like to explain why I'm *not* pushing the fix here rather than leave that looking like an oversight.

Your characterisation is precise: the denominator pin defends against files omitted or dropped, the git census defends against a file nobody imports, and neither requires any individual file to clear a bar. A large thinly-covered file moves the aggregate enough to be caught; small additions and existing slack above 90% are what slip through. My header comment enumerates two vectors and this is a quiet third.

**Not fixing it on this branch, deliberately.** Any push here dismisses your approval — the ruleset sets `dismiss_stale_reviews_on_push` with `required_approving_review_count: 1`, and this PR already lost one approval that way earlier today. Trading a live approval for a change you've explicitly called non-blocking is the wrong direction, particularly on a train where human review is the only thing every open PR is waiting on.

The follow-up also can't be cut from `main` yet, because `ash-typescript-ci.yml` exists only on this branch. So the sequencing is: land this, then the per-file floor goes in against `main` where it can be reviewed on its own merits.

**What that follow-up will carry**, so it's on the record rather than in my head: a jest per-glob `coverageThreshold` alongside the global, along the lines of `"./src/**": {statements: 85}`. Two things I'd want to get right in it rather than now — picking a floor that doesn't immediately fail on files that are legitimately thin (a config shim, a generated barrel), and deciding whether it applies to all files or only newly-added ones, since retrofitting a floor across an existing tree is a different change from gating new code. Both are policy questions this PR shouldn't answer by itself.

PR #501, opened 2026-08-31
fix(scan-set): apply nested ignore files to the subtree they sit in

awsmadi:
## What this changes, and it changes scan results

Nested `.gitignore` and `.ashignore` files have never been honored, on any platform. Their rules
compiled against a synthetic base and matched nothing, so files a user believed they had excluded were
still scanned. This makes them work.

**That means ASH scans fewer files than before in some repositories.** Measured, on Linux, with a
vendored directory that ships its own ignore file:

```
tree: app.py, vendor/.gitignore (contents "*.py"), vendor/lib/dep.py

WITHOUT the fix:  ['app.py', 'vendor/.gitignore', 'vendor/lib/dep.py']   <- dep.py scanned
WITH the fix:     ['app.py', 'vendor/.gitignore']                        <- dep.py NOT scanned
```

That is correct by `.gitignore` semantics and consistent with how a root-level `.gitignore` already
behaves here, but for a security scanner a quiet reduction is the worst possible outcome, so the
reduction is reported rather than silent. See the logging section below.

## The mechanism

`get_ash_ignorespec` received only the collected marker lines, so it had no access to the real scan
root and rebuilt each rule's base from a synthetic `"/"` prefix. A rule from `sub/.gitignore` therefore
compiled against `/sub` while the paths it was matched against live under the scan root, and matched
nothing.

Rules from the *root* ignore file appeared to work, but by accident: the base `"/"` compiles to
`//**/<pattern>`, and the leading `**` swallows any prefix.

The fix threads the scan root in and derives every base from it. There was already in-repo precedent
doing this correctly at `get_scan_set.py:90`, where `parse_rule_file` is called with
`base_dir=root_path`.

Post-fix behavior was checked against `git check-ignore` itself across a range of trees, including a
vendored npm package, a `.venv/.gitignore` containing `*`, an `infra/.gitignore` with `.terraform/`
and `*.tfstate`, a root allow-list of `*` plus `!src/`, and a nested allow-list. ASH and git agree in
every case.

## Two pre-existing bugs this repairs as a side effect

Neither was previously reported, and both are root-level rather than nested:

- **A root `.gitignore` of `*` plus re-includes produced an empty scan set.** The synthetic base made
  `//**/*` match every ancestor path, so nothing survived.
- **Root-level anchored patterns never applied at all.** `/secrets.txt` compiled to `//secrets.txt`
  and matched nothing, so a file the user had explicitly excluded was scanned.

## ASH_INCLUSIONS still wins, which needed a deliberate fix

`ASH_INCLUSIONS` carries `!**/*.template.json` specifically to force CloudFormation templates into the
scan set, overriding git. Making nested ignores work exposed a conflict: git's rule, which `igittigitt`
implements, is that a negation cannot re-include a path underneath an excluded **directory**. So a
`cdk/.gitignore` containing `out/` would have removed `cdk/out/Stack.template.json`, and both
`cdk_nag_scanner.py` and `cfn_nag_scanner.py` consume that scan set, meaning template analysis would
have stopped with nothing logged.

`ASH_INCLUSIONS` negations now win over ignore rules. Verified to be surgical rather than blunt:

```
cdk/.gitignore contains "out/"
  cdk/out/Stack.template.json  KEPT      <- forced back in by ASH_INCLUSIONS
  cdk/out/manifest.json        EXCLUDED  <- correctly, it is not named by ASH_INCLUSIONS
```

## Ordering, because it became load-bearing

`igittigitt` lets the last matching rule decide, and git gives precedence to the ignore file closest to
the file being tested, so ignore files are now read shallowest-first and deeper rules are appended
last. `igittigitt`'s own `parse_rule_files` sorts the same way, so this follows the library's
convention rather than inventing one.

This also fixes a nondeterminism that was previously invisible. The old code built its list with
`list(set(...))`, and a set of strings iterates in hash order, which is randomized per process. A
`!keep.tmp` in `sub/deep/.gitignore` would have beaten `*.tmp` from `sub/.gitignore` on some runs of
the same scan and lost on others. That could not be observed before, because nested rules matched
nothing at all.

## The reduction is reported

Each ignore file logs how many files its rules removed, attributed per file so it is actionable:

```
${SOURCE_DIR}/cdk/.gitignore    excluded 1 file
${SOURCE_DIR}/vendor/.gitignore excluded 2 files
ASH_INCLUSIONS                  excluded 0 files
```

Counts, not file lists, because this runs on every scan and a list of paths per ignore file would bury
the useful line. Zero is emitted rather than omitted, so "nothing was excluded" is distinguishable
from "no nested ignore files were found" — those are different facts and silence would mean both.

The count agrees with the scan set: in the example above `cdk/.gitignore` is credited with 1 and not 2,
because `Stack.template.json` was re-added by `ASH_INCLUSIONS` and is therefore not excluded. A count
that disagreed with the artifact it describes would be worse than no count.

## Two more defects fixed

- **A derived rule base could escape the scan root.** Neither `root / parent_dir` nor a marker's own
  parent was checked for containment, and `igittigitt` normalizes `..` through `os.path.abspath`, so a
  marker naming `${SOURCE_DIR}/../elsewhere/.gitignore` produced a rule rooted outside the tree. The
  base is now confined, with a warning on fallback rather than a silent one.
- **An absolute or `../`-relative `--ignorefile` was silently discarded.** `os.path.join` drops its
  first argument when the second is absolute, so the rules ended up scoped outside the tree and never
  matched. An explicitly supplied ignore file is now applied from the scan root, and says so.

## Verification

- 31 new tests across four files. Full suite **4547 passed / 114 skipped / 0 failed**, identical
  serially and under `-n auto`; the baseline on this branch point is 4516/114/0.
- Every fix has a test that fails against the unfixed code and passes after. Reverting
  `get_scan_set.py` to the base commit fails 3 of the 7 nested-ignore tests.
- Mutation-checked rather than assumed: inverting the ignore-file sort order fails the
  precedence tests, so the ordering claim is pinned rather than incidental.
- `ruff check` on `get_scan_set.py` goes from 33 pre-existing findings to 32; `ruff format --check`
  clean.
- The tests are behavior tests through `scan_set` rather than assertions about compiled rule strings,
  deliberately: the pre-existing tests for this code passed only because they fed synthetic subjects
  rooted at the same fake base the rules used, which is a fixture agreeing with the bug.

## One note for maintainers

`get_scan_set.py:164` compares `Path(path).as_posix()` against a path that `os.walk` joined with the
platform separator, so on Windows the `${SOURCE_DIR}` token is never substituted and the persisted
`ash-ignore-report.txt` records absolute machine paths. This PR handles that case rather than fixing
it, because normalizing the subject of that substitution was measured to break relative scan roots on
Linux: `Path("./sub/.gitignore").as_posix()` drops the leading `./`, which loses the placeholder
entirely, and for `./.gitignore` the anchor then matches the filename's own dot. Comparing
`Path(ignorefile).as_posix()` instead of the raw string looks like the real repair, and would reduce
the absolute-marker handling here to defensive code.


nikmeiser (dismissed):

awsmadi:
Merged `main` in to clear the conflict. This dismissed the approval automatically — the branch ruleset has `dismiss_stale_reviews_on_push`, and there is no way to resolve a conflict without a push. Sorry for the round trip; the change itself is unchanged.

One file conflicted, `utils/get_scan_set.py`, in two hunks:

- **`${SOURCE_DIR}` marker.** `main` added a "KNOWN DEFECT, deliberately not repaired here" comment above the old `re.sub`, closing with *"the real repair is to pass the true scan root into `get_ash_ignorespec`, which also fixes nested ignore files."* That is what this branch does, so the resolution keeps `_source_dir_marker` and removes the comment — it would otherwise warn about a defect the adjacent line fixes. I checked that `_source_dir_marker` covers each failure the comment names (the Windows `os.walk` backslash join, and the reverted normalization that emitted `${SOURCE_DIR}gitignore`).
- **Inclusion loop.** Kept this branch's flat guards. `main`'s inline `"/node_modules/aws-cdk" not in ...as_posix()` is the same test `_is_bundled_cdk_path` performs, so nothing `main` fixed is lost. The two orderings agree on every input except a file matching both the ignore spec and the forced-inclusion spec, where this branch includes it — the behavior the PR is for.

The conflict markers understated the work: this branch adds a required `source_dir` argument to `get_ash_ignorespec`, and `tests/unit/utils/test_get_scan_set_coverage.py` merged cleanly while still calling the single-argument form in ten places. Those call sites now pass `"/"` for the six marker-base tests and `str(tmp_path)` for the four that scan a real tree. **No assertion in that file changed** — its own docstring notes those cases pass because rules and subjects share a synthetic `/` root, so `"/"` keeps them asserting what they did before.

Full suite: 5776 passed, 114 skipped, 3 xfailed. The 31 `ruff` findings in `get_scan_set.py` are all pre-existing on `main` and untouched; the merge happens to drop one `SIM102`.

Left for a separate change: the module docstring and five Windows `skipif` markers in that test file describe defects this branch repairs, so they are now stale. Removing the skips could surface real Windows behavior that deserves its own measurement rather than being folded into a conflict resolution.

nikmeiser (approved):

PR #508, opened 2026-09-02
feat(nix): add --mode nix so scanners come from a pinned flake

awsmadi:
## Why

ASH needs ten external scanner binaries spanning Python, Go, Ruby and Node. Today they arrive baked into a container image or already present on the host PATH, and neither serves an adopter well.

No container image is published, so `--mode container` obliges every adopter to build one first. And `--mode local` degrades quietly: a scanner whose binary is absent reports `MISSING`, contributes zero findings, and the run still writes a complete-looking report. So a scan can under-report without saying so anywhere a reader would notice.

This adds `--mode nix`, which supplies every scanner from a flake pinned by hash. It gives the container's reproducibility without an image or a registry, and it produces no artifact to publish.

## What changed

- `flake.nix` and `flake.lock` pin all ten scanners across `x86_64-linux`, `aarch64-linux`, `x86_64-darwin` and `aarch64-darwin`.
- `nix/opengrep.nix` packages opengrep, the one scanner nixpkgs does not carry.
- `RunMode.nix` plus `automated_security_helper/interactions/run_ash_nix.py` run the scan inside `nix develop`. The shape mirrors container mode: ASH re-executes itself in the shell and the inner run is an ordinary local scan. It is simpler in one respect, since a development shell changes `PATH` but not the filesystem, so none of container mode's path mapping is needed.
- A `nix` entry in the install-method CI matrix, on Linux and macOS across x86-64 and ARM.
- An installation-guide section, written into the `.md.template` since the generated `.md` is regenerated.

## Result

Measured on one fixture, same host and tree:

| | `--mode local` | `--mode nix` |
| --- | --- | --- |
| scanners `MISSING` | 4 (cdk-nag, cfn-nag, grype, syft) | 1 (cdk-nag) |

cdk-nag is `MISSING` in both because its scanner sets `self.command = "python"` and runs from ASH's own environment, so the flake never supplies it. The other three go from absent to running.

## Decisions worth reading before changing this

**Permitting an insecure dependency.** checkov depends on `ecdsa`, which nixpkgs marks insecure for CVE-2024-23342, a timing side channel on P-256. Left alone checkov does not warn, it refuses to evaluate, and the scanner goes `MISSING`. It is permitted through `allowInsecurePredicate` keyed on the package name rather than `permittedInsecurePackages` keyed on `python3.14-ecdsa-0.19.2`, because that string embeds two version numbers and any nixpkgs bump to either silently breaks checkov again. The exposure is narrow: the CVE affects signing, key generation and ECDH, verification is explicitly unaffected, and checkov only reads and evaluates templates. There is also no fixed release to wait for, since affected versions are `>= 0` and upstream considers side channels out of scope.

**Wrapping opengrep's release binary rather than building from source.** This is about parity, not effort. ASH already downloads exactly this asset at runtime, so pinning it keeps Nix mode byte-identical to what the other modes execute. A from-source build would be a different binary, and a cross-mode comparison of finding sets could then diverge for reasons unrelated to the flake.

The published asset is not the scanner. It is a self-extracting launcher that unpacks a 150 MB tree to `$HOME/.cache` on first run, including `opengrep-core`, the OCaml engine. Shipping it as-is fails twice: it needs a writable `HOME` on every run, and the payload it extracts is never patched, so it dies with a loader error on any host without `/lib64`. The derivation extracts once at build time and patches all 64 ELFs, shipping a read-only scanner. Three properties were measured before relying on them: the entry point runs standalone, it runs with the tree read-only, and it runs under a `HOME` other than the one it was extracted under.

**Unsetting `PYTHONPATH` in the shell.** Four scanners here are Python packages, so nixpkgs' Python setup hook appends their whole dependency closure to `PYTHONPATH`, roughly 170 store paths including a pydantic built against python3.14. ASH runs on a different interpreter and inherited that, so Nix's extension modules shadowed its own and the inner scan died on `import pydantic_core._pydantic_core` before doing any work. Unsetting is safe because nixpkgs bakes import paths into the wrapped executables; all four scanners still report their versions afterwards.

**Seeding grype's database.** grype reported `ERROR` rather than `MISSING`, so the binary was fine and only its data was absent: `failed to load vulnerability db: database does not exist`. `ASH_OFFLINE` stops scanners installing tools that would shadow the pinned ones, and the same flag stops grype fetching its database. The container image handles this by running `grype db update` at image build time; a shell has no build step, so it seeds on first entry and says so on stderr. That database changes daily and is far too large to pin in the flake.

**Forwarding argv verbatim.** Only `--mode` is rewritten; the rest of the caller's arguments pass through unchanged rather than being rebuilt from parsed options. Rebuilding is how a forwarded invocation quietly loses a flag. Container mode has to rebuild because its paths need remapping; this does not.

**Refusing rather than falling back.** If Nix is absent the mode fails with an actionable message instead of silently using local mode. A silent downgrade is what produces `MISSING` scanners under a clean-looking report, which is the failure this mode exists to remove.

## Platform support

Linux and macOS, x86-64 and ARM. Windows users should run ASH under WSL2, which uses the Linux path.

Native Windows Nix is not supported, and this is an upstream limitation rather than a gap here. It has no fixed-output derivations, which is the mechanism by which these tools are fetched, and no build sandbox. Nix's own installation documentation lists Windows as WSL2 only.

## Verification

- `nix flake check` passes.
- 23 unit tests pass, with 100% statement coverage of the new module.
- An end-to-end `--mode nix` scan completes with 9 of 10 scanners participating.
- Every one of the ten binaries resolves under the exact name ASH looks up via `find_executable`, which is not the nixpkgs attribute name in two cases (`cfn_nag_scan`, and `npm` for npm-audit).
- The tests were falsification-checked: breaking the recursion guard, the argument rewrite, or the no-fallback refusal each fails a test, so none of those three is merely asserted.
- opengrep's own install check runs a real scan against a fixture rule rather than only `--version`, because `--version` does not load `opengrep-core`, the engine being patched.

## Limitations

`nix flake check` was run on `x86_64-linux` only, as the machine used could not build the darwin or aarch64 outputs. Those three systems are pinned but locally unverified, and the CI matrix covers them.

The nix-versus-container comparison of finding sets was not run; local-versus-nix was used instead.

The CI assertion covers the nine scanners the flake supplies and deliberately excludes cdk-nag, since ASH supplies that from its own environment and its status says nothing about whether the flake worked.


awsmadi:
Merged `main` in (11 commits, no conflicts) and fixed the self-scan failures, which turned out to be this branch's own doing.

**The self-scan legs were failing on `nix/opengrep.nix`.** Pulling the `ash_output` artifact from the failing `scan (python-local, ubuntu-24.04-arm)` leg: 21 detect-secrets findings, **20 already suppressed**, and the 4 that were not are all in the new file —

```
nix/opengrep.nix:54  SECRET-BASE64-HIGH-ENTROPY-STRING
nix/opengrep.nix:58  SECRET-BASE64-HIGH-ENTROPY-STRING
nix/opengrep.nix:62  SECRET-BASE64-HIGH-ENTROPY-STRING
nix/opengrep.nix:66  SECRET-BASE64-HIGH-ENTROPY-STRING
```

Those are the per-system `hash = "sha256-…"` pins for the opengrep release assets — one each for x86_64-linux, aarch64-linux, x86_64-darwin, aarch64-darwin. Four unsuppressed findings is exactly the "4 actionable" the gate reported before failing at a MEDIUM threshold.

An SRI hash of a published release asset is close to the opposite of a credential: it is published so anyone can verify the download, there is nothing to rotate, and it cannot be rewritten to dodge the detector because its value is fixed by the content it pins. So this is a suppression, scoped to the one file rather than a glob over `nix/`, mirroring the entry directly above it that suppresses the same rule on `pyproject.toml` for the same reason.

**Correcting an earlier call on this PR.** I previously reported these failures as not belonging to the branch, on the grounds that no failing job name contained "nix". That was the wrong discriminator — this branch changes what the scan legs *scan*, not what they are *called*. `nix (ubuntu-24.04-arm, py3.12)` itself passed, which is exactly why a name-based check saw nothing. The control that settles it is #494, a sibling PR at the same base with a green rollup: all nine failing job names pass there.

**Verification.** The four findings were replayed through ASH's own `should_suppress_finding` against the parsed config — all four suppressed. Five negative controls in the same pass all stay unsuppressed: a different rule at the same path, an AWS-key rule at the same path, the same rule at `nix/default.nix`, the same rule at `flake.nix`, and the same rule in application code. Without those, a matcher that suppressed everything would have read as success.

Full suite on the merge result: 5813 passed, 114 skipped, 3 xfailed, coverage 93.30%. `uv.lock` was regenerated — git auto-merged it into a state `uv lock --check` rejects, onto versions the merged `pyproject.toml` already requires.

The scan legs need a CI run to confirm; every failing leg runs ASH over the repo, so removing the four actionable findings should clear all of them, but that is a prediction until the matrix reports.

nikmeiser (approved):

PR #514, opened 2026-09-02
fix(cdk-nag): migrate to cdk-nag 3.x and report per-target failures

awsmadi:
## Why

The `[cdk]` extra requires `cdk-nag>=3.0.2,<4.0.0`. The wrapper was written against the 2.x API and cannot call that major, so cdk-nag raised on every template. Separately, those per-template failures were recorded in a local list that was never read, so a run in which every template failed still produced an empty result set and reported success.

Both are fixed here. The second fix is the more general one and is implemented so that it applies to every scanner, not just this one.

## cdk-nag 3.x migration

Two independent 2.x assumptions, each measured against cdk-nag 3.0.2 rather than inferred:

| 2.x | 3.x | Result |
| --- | --- | --- |
| `NagPack(reports=True, report_formats=[...])` | `NagPackProps` has only `verbose`, `writeSuppressionsToCloudFormation` | `TypeError: ... unexpected keyword argument 'reports'` |
| `Aspects.of(stack).add(pack)` | packs implement `IPolicyValidationPlugin`, not `IAspect` | registration attaches nothing; `hasattr(pack, "visit")` is `False`, `hasattr(pack, "validate")` is `True` |
| per-pack `*-NagReport.json` files | one `validation-report.json` via CDK policy validation | globbing for the old files finds nothing |

Packs are now registered with `Validations.of(app).add_plugins(...)` and findings are read from `pluginReports[].violations[]`.

**The SARIF mapping is deliberately untouched.** Under `CfnInclude` a violation's `constructPath` is `<stack>/<template>/<LogicalId>`, so its last segment is the template's logical ID — the same property 2.x's `resourceId` had, which the resource lookup and template line-number search depend on. Verified against a synthesized template, because if it did not hold the lookup would match nothing and drop every finding while appearing to work.

Two details that would each have been a quiet defect:

- Synth **raises** when a plugin reports violations, so the raise is caught and the report read regardless. Letting it propagate would turn every non-compliant template into a scanner error.
- Severity is normalized on the way in. The report emits `"error"` lower-case while the SARIF mapping compares against `"Error"`; passing it through unchanged would classify every finding as a warning.

One behavior difference cannot be preserved: the validation report contains violations only, so `include_compliant_checks` has nothing to include.

## Per-target failure reporting

`cdk_nag_scanner` kept a local `failed_files` list, appended to on both failure paths and never read. Replaced with counters that reach the status decision:

- `ScanResultsContainer` gains `targets_attempted` / `targets_failed`, a `record_target_failure()` recorder, and a `scan_succeeded` property.
- `determine_status()` returns `ERROR` when a scanner attempted targets and failed all of them. This is checked **ahead of** the severity gate on purpose: everything below reasons about finding counts, where zero means "nothing to report", and for a failed scan zero means "nothing was examined" — the same number with the opposite meaning.
- SARIF `executionSuccessful` is derived from that rather than hardcoded `true`.
- Failure logging moves from `trace`, which sits below `debug` and so was invisible even with `--debug`, to `error`.

`ERROR` rather than `FAILED`: `FAILED` means the scanner ran and found problems, which a consumer may legitimately gate or waive on. This did not run.

**The guard is generic.** `ScannerExecutor` copies both counters onto the container alongside `exit_code`, so any scanner reporting per-target outcomes gets it. Scanners that report neither leave both at zero and are unaffected, making this additive for every existing scanner.

## Tests

`tests/unit/utils/test_cdk_nag_wrapper_regression.py` asserts on `inspect.getsource()` **string content** — for example `assert "except (ImportError, FileNotFoundError)" in source`. A test of a function's source text cannot observe whether an API it calls still exists. Measured coverage of `cdk_nag_wrapper.py` from those nine passing tests is **11%**, with the pack construction inside the uncovered range.

Added:

- `tests/integration/scanners/test_cdk_nag_real_pack.py` — constructs a real pack against the installed cdk-nag and asserts non-zero rule and finding counts on a non-compliant template. Confirmed to fail before this change and pass after.
- `tests/unit/models/test_scan_results_container_failure_guard.py` — the status invariant, including the negative cases: a partial failure must **not** report `ERROR`, and a scanner reporting no counters keeps its existing behavior.

Also covers `_non_negative_int_attr`, which exists because `getattr(mock, name, 0)` returns a `MagicMock` rather than the default — mocks auto-create attributes, that value reaches an `int` field which is not validated on assignment, and the first comparison raises. Wiring the guard without that coercion broke 19 existing tests; the coercion and its test are both in this PR.

## CI

Two gaps meant no test could have caught this:

- `uv sync` installs no optional extras, so cdk-nag was absent from every test run. Now `uv sync --extra cdk`.
- `tests/conftest.py` skips anything whose keywords include `integration`, which pytest picks up from the `tests/integration` directory name — so all 15 integration test files have been skipped in CI, and a skipped suite exits 0. A new step runs the cdk-nag integration tests with `--run-integration`, and `ASH_REQUIRE_CDK_EXTRA=1` turns a missing extra into a hard failure rather than a skip.

The new step is **scoped to the cdk-nag file** on purpose: enabling `--run-integration` repository-wide currently surfaces **35 failures across 89 tests** in unrelated integration tests. Repairing and enabling the rest of that suite is worth doing and is not in scope here.

## Verification

- Full suite: **4534 passed, 117 skipped, 0 failed**.
- The new integration tests fail against the unmigrated wrapper and pass after — checked in that order.
- `uv.lock` is deliberately untouched. It currently drifts from `pyproject.toml` for unrelated dependencies, and folding that into this PR would mix concerns.

## Breaking change

A scanner that attempted at least one target and failed on all of them now reports `ERROR` with a non-zero exit code and SARIF `executionSuccessful=false`, where it previously reported `PASSED` with zero findings. Pipelines running a scanner that cannot execute will begin failing rather than completing. The `BREAKING CHANGE:` footer is in the commit so commitizen renders it into the changelog on the next release bump.


awsmadi:
Merged `main` in and brought the test suite onto the v3 interface. Full suite is **5827 passed, 117 skipped, 3 xfailed**, coverage 93.32%.

27 tests were failing at `from aws_cdk import Validations` with `cannot import name ... (unknown location)`. That suffix distinguishes two very different causes: Python names a file path there for a real package and says "unknown location" for a module object without one. So this was `tests/unit/utils/test_cdk_nag_wrapper_behavior.py`'s hand-built `types.ModuleType` double, not a missing symbol in the pinned dependency — aws-cdk-lib 2.267.0 does export `Validations`, and a real synthesis of a non-compliant template through the migrated wrapper returns 3 findings at `level=error`, `kind=fail`. The doubles were faithful to 2.x; the interface moved.

Porting them rather than deleting them turned up three defects in the migration:

| defect | fix |
| --- | --- |
| An unresolvable pack name was logged and skipped, so a request for three packs could evaluate two and exit zero. The total-failure guard missed it because *some* packs registered. | Raises before anything is synthesized. This is the partial-scan-reports-whole shape this PR exists to remove, reintroduced one level up. |
| `validation-report.json` holding literal `null` decodes to `None` without error, so the `try/except` around the parse missed it and the next `.get` raised `AttributeError` out of the scanner. 2.x had this case covered. | The decoded type is checked. |
| The level/kind mapping was two nested conditionals inside the `Result` construction, reachable only by a full synthesis. | Extracted as `_level_and_kind` and tested directly. |

The extraction matters because the v3 report carries **violations only**: three of the mapping's four rows cannot arise from a real scan, so a test faking a report to reach them would assert against a payload cdk-nag cannot write.

**On the doubles.** They are copied from a report emitted by the installed aws-cdk-lib and cdk-nag, not from the wrapper's parser — doubles written from the parser agree with it by construction and say nothing about whether it reads what cdk-nag writes. Three deliberate details:

- `violatingResources` is in the fixture, set to `null`, beside `violatingConstructs`. The real report carries both that way, while the jsii class exposes only the snake_case `violating_resources`. A parser written from the class reads the null key and reports a violating template as clean; the decoy makes that fail a test.
- Severity arrives lowercase. Compared raw against `"Error"` it demotes every finding to a warning and slips it under a severity gate. Casing is covered, and an unrecognized severity is treated as an error rather than downgraded.
- `Aspects` is absent from the `aws_cdk` double, so a regression to the 2.x aspect registration fails at import instead of attaching nothing and evaluating no rules.

**Two scanner tests were asserting on the ambient environment.** `test_missing_cdk_dependencies_fail_validation` and `test_unsatisfied_dependencies_return_false` got the state they needed by accident — they asserted that validation fails while relying on `_CDK_AVAILABLE` being `False` from an import that failed at module load. That holds only while nothing installs the `cdk` extra, and this branch's unit-test action runs `uv sync --extra cdk` so the integration tests can execute. Both now take a `cdk_unavailable` fixture that forces the flag, mirroring the `cdk_available` fixture that already did this properly.

**Verification.** Beyond the suite: the 3 real-pack integration tests pass under `--run-integration` with `ASH_REQUIRE_CDK_EXTRA=1`, which turns their skip into a hard failure. Then every load-bearing line was broken on purpose to check the tests notice — 7 mutations, 7 killed, baseline green before and after. Reading `violatingResources` kills 22 tests; registering plugins on the stack kills 32; taking the first construct-path segment instead of the last kills 21; skipping severity normalization kills 6. The harness asserts each replacement actually changed the file and that the restore is byte-for-byte, since a find-and-replace matching nothing would otherwise report a clean sweep having tested nothing.

Two notes for review:

- **`uv.lock` was regenerated.** Git auto-merged it into a state matching neither side; `uv lock --check` failed on the merge result and passes on both parents. The regeneration moves it onto versions the merged `pyproject.toml` already requires after the dependency bumps that just landed.
- **No CHANGELOG entry.** That file is generated by commitizen at bump time, so a hand-written entry would be overwritten. The breaking change is in the commit's `BREAKING CHANGE:` footer, which is what feeds it — covering both the raise on an unknown pack name and `include_compliant_checks` no longer having any effect, since the v3 report has no compliant or suppressed records at any setting.

automated_security_helper/utils/cdk_nag_wrapper.py
```diff
@@ -196,20 +387,47 @@ def get_nag_packs():
...
-                Aspects.of(stack).add(pack_instance)
+                return None
 
-            app.synth()
+            # Synth is where validation runs, and CDK raises when a plugin reports violations.
+            # For this wrapper a raise is the ordinary case -- findings are the product -- so
+            # it is caught and the report is read regardless. Letting it propagate would turn
+            # every non-compliant template into a scanner error.
+            try:
+                app.synth()
```
nikmeiser, line 463: Your `test_a_missing_report_is_reported_as_nothing_scanned` docstring says the failure is carried "in the scanner's target-failure counters," but `cdk_nag_scanner.py` only bumps `targets_failed` on an exception (`:325`); a missing report returns a non-None empty response (`cdk_nag_wrapper.py:565`) and never raises, so targets_failed stays 0 and the scan passes (`scan_succeeded` true even for N=1). Either raise from the wrapper on a missing report so the scanner's `except` fires, or have the scanner count an empty/errored response as a failed target — and add a scanner-level test that a report-less template fails the scan.

automated_security_helper/plugin_modules/ash_builtin/scanners/cdk_nag_scanner.py
```diff
@@ -272,13 +272,20 @@ def scan(
...
+        # an empty-but-successful report. They are attributes rather than locals precisely so
+        # the executor can read them and status computation can see them.
+        self.targets_attempted = 0
+        self.targets_failed = 0
         target_rel_path = get_shortest_name(input=target)
 
         outdir = self.results_dir.joinpath(target_type)
         sarif_results: List[Result] = []
         for cfn_file in scannable:
+            self.targets_attempted += 1
```
nikmeiser, line 290: In the scanner (`cdk_nag_scanner.py:288–336`): `targets_attempted += 1`; `nag_result_dict is None` is
false (it's an empty response, not `None`); `.results.items()` is empty so no findings; no exception is
raised, so the `except` block that does `self.targets_failed += 1` never runs.

The `None`-vs-`{}` split in `_violations_from_validation_report` and the "NOT scanned" ERROR log sounds right, and the intent is that a report-less synth crash should fail the scan, not pass silently. 

It doesn't look like the intent is actually realized. `test_a_missing_report_is_reported_as_nothing_scanned` says the distinction "has to live in the log **and in the scanner's target-failure counters**" — and that second half doesn't hold in `cdk_nag_scanner.py`:
- `_violations_from_validation_report` returns `{}` on a missing report (`cdk_nag_wrapper.py:180-185`) and the wrapper returns a **non-None** `CdkNagWrapperResponse(results={}`) (`:565`) — it never raises.
- The scanner only increments `targets_failed` inside its `except` block (`cdk_nag_scanner.py:325`). A **non-None** empty response isn't `None` (so it's not skipped) and doesn't raise (so `except` never runs) — it falls through the `for … in .results.items() `loop with zero iterations.
- So `targets_failed` stays `0`, and `scan_succeeded = (targets_attempted <= 0` or `targets_failed < targets_attempted)` is `0 < 1` = **True**. A template whose synth crashed without writing a report is counted as a clean scan — and this passes **even for a single-template repo**, not just partial failures. The "failed on all N templates" guard (`:334-340`) can't catch it either, since it needs `targets_failed == targets_attempted` and a missing report produces zero failures.

The wrapper deliberately makes a three-way distinction (`None` = skip / `{}` = ran-but-empty / raise), but the scanner only reads two of them (`None` vs raised), so the `{}` signal gets dropped.

Two ways to close it, either works:
1. Raise from the wrapper on a missing/unparseable report after synth, so the scanner's existing except path fires and increments targets_failed; or
2. Have the scanner treat an errored/empty response as a failed target (e.g. `if not nag_result_dict.results: self.targets_failed += 1`).

Either way, worth adding a scanner-level test — the current assertion is in `test_cdk_nag_wrapper_behavior.py` and only checks `results == {}` + the log, so nothing pins that a report-less template actually fails the scan. That's the gap that let the docstring and the code diverge.

awsmadi: Confirmed and fixed in c1152a5b. Your trace was exact — non-None response, no raise, zero-iteration loop, `targets_failed` stays 0, and `0 < 1` reports success even with a single template.

I took a third option rather than either of the two you offered, because both infer the outcome. `CdkNagWrapperResponse` now carries `failure: str | None`, set to a reason string whenever `_violations_from_validation_report` could not read a report, and the scanner branches on that field. Raising would collapse the deliberate skip/ran/failed distinction into one channel; `not response.results` works only because a compliant pack yields `{pack: []}` rather than `{}`, and an unasserted invariant of exactly that kind is what produced this defect. The reason string also reaches `self.errors`, which a boolean couldn't.

One case fell out of it that wasn't handled before: a report that parses but lists no plugin reports. Every registered pack writes an entry even when it finds nothing, so an empty mapping means no pack reported.

**On the docstring** — you're right that it's where the divergence lived. It claimed the distinction was carried "in the scanner's target-failure counters" while only testing the log. It now states what the test checks and points at the scanner-side half.

**Scanner-level tests added**, as you asked: a report-less template is a failed target, the reason reaches `self.errors`, and with one template the scan is not successful. Plus two negative controls — a compliant `{pack: []}` is not a failed target, and a non-CloudFormation `None` is still neither attempted nor failed — since a fix counting any empty result as failure would look identical while making every compliant repo a scanner error.

Worth reporting that the first version of those tests wasn't sufficient, and a mutation run caught it. Breaking the scanner branch went red, but hardcoding `failure=None` on the wrapper's return stayed green — the scanner tests drive a wrapper double and never exercise that return, so the wrapper half was unpinned in the same shape you describe one level up. Swapping the explicit field for `not results` also survived, because nothing distinguished them. Both are closed now: a seven-shape parametrized test covers `failure` in both directions, and a scanner test asserts a response carrying `failure` is a failed target even when it also carries findings.

Full suite 5839 passed / 117 skipped / 3 xfailed, coverage 93.33%; 3 real-pack integration tests green; 3 mutations, 3 killed.

awsmadi:
Widening a verification claim I made too narrowly.

Earlier I reported that aws-cdk-lib **2.267.0** exports `Validations`, which is what `uv` happened to resolve here. But the pin is `>=2.266.0,<3.0.0`, and a resolution is not a range — a user installing today gets the newest, while a lockfile, a constraints file, or an older index snapshot can hand them the floor. The boundary is the case that matters, and it was the one I had not tested.

Tested now, each in its own throwaway venv so the versions cannot contaminate each other, with the resolved version printed alongside the result so a symbol found under an unexpected version cannot read as a pass:

| aws-cdk-lib | `Validations` | `IPolicyValidationPlugin` | `PolicyViolation` | `Validations.of` / `.add_plugins` |
| **2.266.0** — the pin's floor | present | present | present | present |
| **2.257.0** — the floor at v3.6.0 / v3.7.0 | present | present | present | present |

So the migration is viable across the entire pinned range rather than at one resolved point, and it needs no floor bump. The 2.257.0 row matters because that is the floor pinned at the released tags, which is where this change would land if it were ever backported — it means the same diff applies there without touching dependency constraints.

Nothing in the change is affected. This only replaces "verified at 2.267.0" with "verified at the pin's floor and above", which is the claim I should have made the first time.

automated_security_helper/utils/cdk_nag_wrapper.py
awsmadi: Answered in the sibling thread on `cdk_nag_scanner.py:290`, but this one raised the docstring specifically, so recording it here rather than leaving it open.

Both asks landed in c1152a5b, which is on this PR.

The mechanism is a third option rather than either of the two you offered, because both infer the outcome from a side effect. `CdkNagWrapperResponse` now carries `failure: str | None`, set whenever `_violations_from_validation_report` could not read a report, and the scanner branches on that field. Raising would collapse the deliberate skip / ran / failed distinction into a single channel, and `not response.results` works only because a compliant pack yields `{pack: []}` rather than `{}` -- an unasserted invariant of exactly that kind is what produced this defect. The reason string also reaches `self.errors`, which a boolean could not carry.

On the docstring, which is what this thread flagged: it claimed the distinction was carried "in the scanner's target-failure counters" while the test only checked the log. It now states what the test actually checks, and points at the scanner-side half.

Scanner-level tests added as you asked: a report-less template is a failed target, the reason reaches `self.errors`, and with a single template the scan is not successful. Plus two negative controls, since a fix that counted any empty result as a failure would look identical while turning every compliant repository into a scanner error.

One thing worth reporting. The first version of those tests was not sufficient, and a mutation run caught it. Breaking the scanner branch went red, but hardcoding `failure=None` on the wrapper's return stayed green, because the scanner tests drive a wrapper double and never exercise that return -- the same shape you describe, one level up. Both are closed now: a parametrized test covers `failure` in both directions, and a scanner test asserts a response carrying `failure` is a failed target even when it also carries findings.

nikmeiser (approved):
Empty packs handling needs to be improved at a future date to display 'skipped' instead of green.

PR #533, opened 2026-09-08
build(deps): declare minor-range floors so patch releases need no manifest edit

awsmadi:
## What

Dependency declarations now name a `major.minor` floor instead of pinning a patch. Three changes that only work together:

1. `pyproject.toml` floors drop their patch component (`requests>=2.34.2` becomes `requests>=2.34`). Upper bounds are untouched.
2. The Python Dependabot entry gains `versioning-strategy: increase-if-necessary`.
3. That entry's `package-ecosystem` changes from `pip` to `uv`.
4. A new entry monitors `ash-agent-plugins/agentic-coding/transpiler/`, which nothing was watching.

## Why

A patch-precision lower bound over-constrains consumers for no benefit. `requests>=2.34.2` already admits 2.34.3, so the patch digit is not what makes new releases usable — it just forbids 2.34.0 and 2.34.1 for no stated reason.

It also generates churn. Dependabot's [documented default strategy](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference) for an application is `increase`, which "always increase the minimum version requirement to match the new version". Every upstream patch release therefore produced a pull request whose entire diff was one digit of a floor. Recent examples: #532, #530, #528, #527, #526, #525, #524, #523.

That churn is not merely noisy. One of those bumps raised `aws-cdk-lib` in `pyproject.toml` while a hand-maintained copy of the same requirement elsewhere stayed behind, and the resulting drift failed nine unit-test legs on an unrelated pull request.

### Why the strategy change is required, not optional

Loosening the floors alone would not survive contact with Dependabot: under the default `increase`, the next patch release re-adds a patch floor to every line, reverting this change within days. `increase-if-necessary` "leave[s] the version requirement unchanged if it already allows the new release (Dependabot still updates the resolved version)", which is exactly the wanted behavior. Out-of-range releases still widen the requirement, so a major is not silently skipped the way `lockfile-only` would skip it.

### Why the ecosystem change is required too

The Python entry declared `package-ecosystem: "pip"`, but this project resolves from `uv.lock`. Of the last 22 Python Dependabot pull requests, **22 changed `pyproject.toml` and none changed `uv.lock`** — the lockfile was never maintained. That also explains why `uv lock --check` was failing on `main`: merged bumps left the lock behind, on six specifiers.

So `increase-if-necessary` on a `pip` entry would have been worse than the status quo. It would leave satisfied floors alone *and* still not touch the lock, freezing Python updates until an out-of-range release. `uv` is a supported ecosystem value, and it supports `versioning-strategy`; every other option already in this file (`groups`, `cooldown`, `commit-message`, `labels`, `schedule`) carries no ecosystem restriction, so the existing configuration survives the move unchanged.

## Four floors deliberately keep their patch component

Truncating a floor lowers it, because PEP 440 zero-pads: `>=3.1` means `>=3.1.0`. For four packages that would re-admit a release with a published advisory, or is meaningless. Each says why inline, so a later cleanup does not silently strip it:

| Package | Floor | Truncating would admit |
| --- | --- | --- |
| `GitPython` | `>=3.1.61` | 3.1.0, and GHSA-wvpp-8hx9-p66j plus three others affect `<= 3.1.57` |
| `mkdocs-material` | `>=9.7.7` | 9.7.0 through 9.7.6; GHSA-xvg9-69gf-fjrf affects `>= 7.2.0, < 9.7.7`, so the floor is the fix version |
| `pymdown-extensions` | `>=11.0.2` | 11.0.0; GHSA-gm37-52c6-37mw affects `<= 11.0.0` |
| `python-multipart` | `>=0.0.32` | 0.0.1 — the package is `0.0.x`, so the patch is the only meaningful component and `>=0.0` says nothing |

`cryptography` was checked and is safe to truncate: its floor existed to pick up GHSA-g6cj-pr64-35w5, which affects `>= 44.0.0, < 50.0.0`, so `>=50.0` preserves that boundary exactly. The `.1` came from a routine bump, not from the advisory.

Every other floor was checked against the GitHub Advisory API for whether the newly admitted window contains a known-affected release. None does.

## Verification

- `uv run pytest tests/unit` — 5813 passed, 22 skipped, 3 xfailed. Coverage 93.30%.
- `uv.lock` regenerated: 47 insertions, 47 deletions, **all of it `specifier =` text** inside `requires-dist` metadata. **No resolved version changed, in either direction.** Loosening a floor only widens what is permitted, so the resolver still selects the newest release; a downgrade here would have meant a floor was load-bearing.
- `uv lock --check` failed on `main` before this change and passes after. The six stale specifiers it was complaining about are fixed as a side effect of regenerating.
- The manifest parses and the wheel builds.

## Notes for reviewers

- Upper bounds are untouched throughout. Normalizing cap style (`<3.0.0` versus `<3`) is deliberately out of scope; mixing it in would bury the change.
- `ash-agent-plugins/agentic-coding/transpiler/pyproject.toml` needed no rewrite. Its floors are already two-component, which is the precedent this change follows.
- The transpiler entry is monitoring only. `directory` scopes a Dependabot entry to one manifest, so the root entry never covered that project.


nikmeiser (approved):
Thanks. Missing is the right decision.

nikmeiser (approved):

PR #534, opened 2026-09-08
feat(plugins): make ASH installable with npx skills, and document agent installs

awsmadi:
## What

Makes ASH installable with Vercel's open agent-skills CLI, and documents the three ways to wire ASH into an AI coding agent.

1. A new transpiler backend emits a repository-root `skills/ash-mcp/` tree from the existing `_base/` source.
2. A CI job gates the generated plugin trees against drift, in ASH's own workflows.
3. `.gitignore` no longer hides the generated Copilot `mcp.json`.
4. The root `README.md` gains an install section; `ash-agent-plugins/README.md` gets corrected.

## Why the root directory

The `skills` CLI discovers skills at a repository-root `skills/<name>/` path. ASH's skill content and frontmatter were already valid, but every copy lived under `ash-agent-plugins/agentic-coding/plugins/<target>/skills/`, so the CLI found nothing. Measured before this change:

```
$ npx skills add <checkout> --list
No skills found
No valid skills found. Skills require a SKILL.md with name and description.
```

And after:

```
$ npx skills add <checkout> --list
Found 1 skill
  ash-mcp
    Run security scans with the ASH (Automated Security Helper) MCP server. ...
```

The tree is generated rather than copied, so it cannot drift from `_base/`. It is byte-identical to the `generic-skill` backend's output (`diff -r`, no differences); the two differ only in where they land. `generic-skill` stays inside the plugin tree for anyone assembling a custom integration.

## How the output root is expressed

Every existing backend wrote under `agentic-coding/plugins/`. Rather than a relative escape in a template, `BaseBackend` gained a declared anchor selector resolved through one funnel:

```python
OutputAnchor = Literal["plugins", "repository"]

class BaseBackend:
    OUTPUT_ANCHOR: ClassVar[OutputAnchor] = "plugins"

def resolve_output_dir(backend_cls, anchors: OutputAnchors) -> Path:
    output_dir = validated_output_dir(backend_cls.NAME, backend_cls.OUTPUT_DIR)
    return anchors.root_for(backend_cls.OUTPUT_ANCHOR) / output_dir
```

All six sites that compute an output directory now go through it. The repository root is found by walking up for a `.git` entry and raising if none is found, rather than a fixed hop count: `ash-agent-plugins/` is designed to be extractable as its own repository, where the transpiler sits two levels below the root instead of three. It tests `exists()` rather than `is_dir()`, because `.git` is a file in a linked worktree.

### Two guards, both proven able to fail

`run_section_emitters` begins with `reset(out)`, which is `shutil.rmtree`. A backend resolving to the repository root would delete the repository. Two safeguards:

- `validated_output_dir` rejects an `OUTPUT_DIR` that is empty, `.`, `..`, absolute, or a UNC root, consulting both `PurePosixPath` and `PureWindowsPath` — `PurePosixPath` does not treat `C:\x` as absolute and does not split on backslash. It runs at registration, so a misdeclaration fails at import rather than mid-build.
- `check_drift` builds into a temporary directory. Its two anchors are independent explicit fields with a sandbox constructor, not derived from one another; deriving the repository anchor from the sandboxed plugins root would resolve two levels above the temporary directory and then rmtree there.

Both were verified by mutation rather than by observing a pass. Sandboxing only the plugins anchor produces:

```
AssertionError: check_drift overwrote the tree it was asked to compare;
its sandboxed build escaped into the repository anchor
```

Weakening the escape guard fails 9 tests. Transpiler tests go from 3 to 26.

## The CI gate, and why one was needed

`ash-agent-plugins/README.md` claimed a pre-commit hook and CI verified that committed plugin files match transpiler output. That holds only when `ash-agent-plugins/` is its own repository. In this repository it was inert: GitHub reads workflows only from the repository root, so `ash-agent-plugins/.github/workflows/validate.yml` never ran, and the root `.pre-commit-config.yaml` has no transpiler hook. All 16 generated trees were unguarded against hand-edits.

`.github/workflows/ash-agent-plugins-drift.yml` runs `agentic-plugins check` plus the transpiler test suite. Its path filters cover the transpiler tree, `_base/`, the generated `skills/` tree, and the workflow file itself — the nested workflow has no filters, so these were written from scratch. Proven to fail: hand-editing the generated `SKILL.md` exits 1 naming `[skills-root] ~ ash-mcp/SKILL.md`; reverting exits 0.

## A prerequisite fix

`agentic-plugins check` was **already failing on `main`**. `ash-agent-plugins/.gitignore` matched `.vscode/`, which made the Copilot backend's generated `.vscode/mcp.json` untrackable, so a clean checkout reported a missing declared path and Copilot drift that no rebuild could clear. Nobody had seen it because the only workflow running that check does not execute here.

This is more than a red gate. On `main` the Copilot tree contains no `.vscode/` at all, so that README's instruction to copy `.github/` and `.vscode/` into your repository was unfollowable — the directory was not in the repository. The Copilot MCP wiring ships for the first time here.

## Documentation

Root `README.md` had no mention that the plugin distribution exists. It now covers the skill, the per-platform plugins, and the MCP server directly, as alternatives rather than steps, with the prerequisite stated and a pointer to the existing MCP section. The 17-row platform table is **not** duplicated; the root README links to it, because a second hand-maintained copy is the drift this repository keeps paying for.

Corrections made while there, all pre-existing:

- The false CI claim, in the three places it appeared.
- Three documented `transpile --check` / `--validate-only` / `--drift-only` invocations. All three fail with `ImportError`: `pyproject.toml` declares `transpile = "transpiler.cli:transpile_compat"`, and that symbol does not exist. Replaced with `agentic-plugins check`, flags verified against `--help`.
- The platform count, now measured from the registry: 15 platform plugins plus 2 agentskills-format releases, 17 outputs total.
- Three table-of-contents entries: one for a heading that does not exist, two headings that were missing.

Two commands were deliberately described rather than scripted. `cp -r .../.cursor ./` nests a second `.cursor` inside an existing one, so it ships a footgun; and `codex plugin marketplace add` registers a marketplace rather than installing a plugin, which the plugins README overstates.

## Known limitation

`npx skills add awslabs/automated-security-helper` starts working when this merges. The CLI resolves that shorthand against the repository's default branch, so before merge the documented command reports "No skills found". The local-path form is verified working today. The documentation describes the merged state.

## Follow-ups, not in scope here

- The Architecture section of `ash-agent-plugins/README.md` documents `transpile.py`, `schema.py`, and `configs.yaml`. None of the three exists; the configuration moved into `BaseBackend` class variables. Fixing it means rewriting the directory tree plus prose in two sections that still treat `configs.yaml` as live.
- The broken `transpile` console script entry point.
- `agentic-plugins smoke-test` reports a `codex` version mismatch against the pinned value in `_base/cli_versions.json` on hosts with a newer `codex`. CI installs the pinned version, so the gate here is unaffected.

### DIFF
--- a/.github/workflows/ash-agent-plugins-drift.yml
+++ b/.github/workflows/ash-agent-plugins-drift.yml
@@ -1,0 +1,117 @@
+name: Agent Plugins Drift
+
+# WHY THIS LIVES AT THE REPOSITORY ROOT
+#
+# ash-agent-plugins/ ships its own workflow at
+# ash-agent-plugins/.github/workflows/validate.yml, and ash-agent-plugins/README.md
+# describes it as the gate that keeps the committed plugin files matching
+# transpiler output. That is true only when ash-agent-plugins/ is checked out as
+# its own repository, which it is designed to be. Inside this monorepo the file is
+# inert: GitHub reads workflows only from .github/workflows/ at the repository
+# root, so nothing in that nested directory has ever run here. The root
+# .pre-commit-config.yaml has no transpiler hook either.
+#
+# The cost of that gap was not theoretical. It let a permanent drift failure sit
+# on main undetected: ash-agent-plugins/.gitignore ignored .vscode/, which
+# silently swallowed the copilot backend's generated .vscode/mcp.json, so
+# `agentic-plugins check` reported drift on a clean checkout and no amount of
+# re-running the transpiler could clear it.
+#
+# This workflow is the same check, invoked from the root with
+# ash-agent-plugins/agentic-coding/transpiler prefixes so it runs in ASH's own CI.
+# The nested workflow is left alone -- it remains correct for the standalone
+# layout, and the two do not conflict.
+#
+# WHAT THE CHECK ACTUALLY PROVES
+#
+# `agentic-plugins check` rebuilds every backend into a temporary directory and
+# byte-compares against the committed output, then runs schema and structural
+# validation over the result. It fails if a committed plugin file disagrees with
+# what _base/ would generate today -- which is what stops the generated trees from
+# drifting away from their single source of truth. The repository-root skills/
+# tree is covered by the same comparison, so it cannot silently diverge from the
+# generic-skill artifact either.
+#
+# PATH FILTERS, WRITTEN FROM SCRATCH
+#
+# The nested workflow has no path filters at all, so there was nothing to copy.
+# These four cover every input the check reads and every output it compares
+# against:
+#
+#   - the transpiler itself, which includes _base/ (the source of truth),
+#     templates/, schemas/ and tests/
+#   - the generated per-platform plugin trees, the right-hand side of the
+#     byte comparison
+#   - the generated repository-root skills/ tree, likewise
+#   - ash-agent-plugins/.gitignore, because a pattern added there can make a
+#     generated file untrackable and reintroduce exactly the permanent drift
+#     described above -- an edit to that one file, alone, must run this gate
+#
+# This workflow file is in its own filter. Unlike ash-typescript-ci.yml, which
+# deliberately omits itself because its target directory does not exist on the
+# default branch yet, everything this gate checks is present, so the change that
+# edits the gate is tested by the gate.
+#
+# Consequence to be aware of: a change outside these paths cannot break the
+# plugin trees, but it also will not run this job.
+
+on:
+  push:
+    branches:
+      - "*"
+    tags:
+      - "*"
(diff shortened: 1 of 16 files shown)


nikmeiser (approved):
- `.github/workflows/ash-agent-plugins-drift.yml:127` use SHA instead of version tag (supply chain attack protection)
- `ash-agent-plugins/agentic-coding/plugins/copilot/.vscode/mcp.json:6` hardcoded to ash@3.4.0. Current version is 3.7.1

PR #542, opened 2026-09-09
fix(plugins): stop the installed ASH tag going stale, and SHA-pin two actions

awsmadi:
Addresses the two review comments left on #534. Both were valid, and both were landed unfixed when #534 merged, so they are defects on `main` now rather than outstanding PR feedback.

## Warning for anyone about to cut a release

Two version facts that interact badly. Neither is fixed here, and both are listed under follow-ups at the bottom.

`cz bump` currently computes **3.7.0 -> 4.0.0**. A breaking change has accumulated since the last bump, so the next person who runs it cuts a major release rather than a patch. Confirmed with `cz bump --dry-run`, which prints `bump: version 3.7.0 -> 4.0.0` and `tag to create: v4.0.0`.

Separately, `pyproject.toml` says `3.7.0`, but **v3.7.1 is the published Latest release** (2026-09-03) while **v3.7.0 is still a Draft**. A release shipped without the bump landing on `main`. Both git tags exist.

If you reconcile that version, move `[tool.commitizen] version` and the four `version_files` targets added here **together**. Bumping `pyproject.toml` alone puts those four files ahead of the configured version, which recreates the exact silent-staleness bug this PR fixes, for the reason explained under "Why v3.7.0" below.

## Why this is a follow-up and not a push to #534

#534 squash-merged as `3eac29d8` and its head branch `feat/npx-skills-root-target` was deleted, so there is no branch left to push to. Its `headRefOid` is frozen at `48e71e9e`, so a push could not move the review either. This branch is based directly on `main`.

Worth noting for anyone comparing: `48e71e9e` is not an ancestor of `main`. It is behind `main` by #533 and #535, so building on the old PR head would have reverted those and deleted three test files.

## Issue 1: pin the workflow's actions by commit SHA

`.github/workflows/ash-agent-plugins-drift.yml` named `actions/checkout@v7` and `astral-sh/setup-uv@v7`. A tag is a mutable pointer, so whoever controls the action repository can move `v7` to any commit and this job runs it with no diff here to review. That matters for these two steps in particular: checkout handles the repository token, and setup-uv populates the dependency cache the job then builds from.

| Action | Tag | Tag object | Pinned commit |
| --- | --- | --- | --- |
| `actions/checkout` | v7 -> v7.0.1 | lightweight | `3d3c42e5aac5ba805825da76410c181273ba90b1` |
| `astral-sh/setup-uv` | v7 -> v7.6.0 | annotated | `37802adc94f370d6bfd71619e3f0bf239e1f3b78` |

The annotated tag is the part to be careful about. `astral-sh/setup-uv`'s `v7` ref names a tag **object** (`94527f2e458b27549849d47d273a16bec83a01e9`), not a commit. That object was dereferenced to the commit above; pinning the tag object's own SHA would name something a runner cannot check out. `actions/checkout` publishes `v7` as a lightweight tag, so its ref already names a commit and needed no dereference.

Pinning normally trades freshness for immutability, and that trade is not being made. `.github/dependabot.yml` already declares a `github-actions` ecosystem entry for `/` on a daily schedule, and Dependabot reads the trailing `# v7.x.y` comment to learn which release a SHA corresponds to, rewriting the SHA and the comment together. The comments are load-bearing, not decorative.

**Scope is this workflow only.** Sixty-four other external `uses:` references in this repository are still tag-pinned. Those, plus an enforcement gate that fails on any unpinned reference, belong in a separate pull request. Adding the gate here would turn CI red against the sixty-four references this PR does not touch.

## Issue 2: make releases maintain the ASH tag the plugins install

The reviewer cited `plugins/copilot/.vscode/mcp.json:6` pinning `ash@3.4.0`. That file is generated. Editing it would have fixed one copy of eighteen and the next transpiler run would have reverted it. The literal lives in `transpiler/_base/`, which renders into `AGENTS.md`, 17 per-platform plugin trees, and the repository-root `skills/` tree.

### Scope correction: four files, not two

`_base/` carried the tag in **four** files, not the two the report traces to:

- `manifest.json` -- the `ash_version` field, structured data
- `mcp.json` -- the `uvx --from=git+...@<tag>` argument, structured data
- `commands/scan.md` -- the install command surfaced to the user when a check fails
- `references/troubleshooting.md` -- the documented recovery command

The last two hold the tag in prose a user or an agent is told to paste, which is easy to read as cosmetic. It is not. `uvx --from=git+...@<tag>` resolves a **git ref**, so all four fail identically: a stale tag quietly installs an old ASH, and a deleted tag fails outright. A reviewer who checked only `manifest.json` and `mcp.json` would think this PR is over-scoped.

### The mechanism

Nothing compared `_base/` against the package version, so a release could move the version and leave these behind with no failure anywhere. `Formula/ash.rb` already had this exact defect, which is why it sits in `[tool.commitizen] version_files` with a comment explaining it. These four join it. That is the fix: `cz bump` now rewrites them on every release.

### Why v3.7.0 and not v3.7.1

The value written is `v3.7.0`, matching `[tool.commitizen] version`, and not the newer published `v3.7.1`. This is deliberate and it is the load-bearing decision in the PR.

commitizen does `line.replace(current_version, new_version)` on lines matching each pattern -- it searches for the **current** version. A `version_files` target written as `v3.7.1` while `[tool.commitizen] version = "3.7.0"` therefore matches the line and rewrites nothing. The file silently stops being maintained from that point on, which is the defect being fixed rather than a fix for it.

Negative control, simulating `cz bump 3.7.0 -> 3.7.1` against the pre-fix content on `main`:

```
_base/manifest.json                    matched=1 rewritten=0  =>  silent staleness
_base/mcp.json                         matched=1 rewritten=0  =>  silent staleness
_base/commands/scan.md                 matched=1 rewritten=0  =>  silent staleness
_base/references/troubleshooting.md    matched=1 rewritten=0  =>  silent staleness
```

Nothing would have caught that: the release workflow runs `cz bump --changelog --yes` with no `--check-consistency`, so a target that matches but rewrites nothing is skipped without a word. Git tag `v3.7.0` exists, so the install command resolves; the v3.7.0 GitHub Release being a draft does not affect a git ref.

### Correction on how to verify this

`cz bump --dry-run` cannot demonstrate a rewrite. `Bump.__call__` returns before `update_version_in_files`, so the dry run exits 0 having changed **0 files**. `cz bump --files-only` is the valid proof: it rewrites the version files and runs `pre_bump_hooks`, then returns before the commit and tag. Stated here so the next person does not verify it the wrong way and conclude the config is broken.

Running `cz bump --files-only --yes` rewrote all four `_base/` files, propagated to all 30 generated files, left zero of them on the old version, and left `agentic-plugins check` at exit 0. No commit and no tag were created.

### Two gaps closed, each proven load-bearing

**A bump that touched only `_base/` would cut a guaranteed-red gate.** The 17 generated trees stay on the old tag until the transpiler reruns, and `agentic-plugins check` byte-compares them. Reverting just the generated trees while leaving `_base/` bumped makes `check` **exit 1** with `ERROR: generated outputs differ from _base/ source. Run agentic-plugins build and commit the result.` A `pre_bump_hooks` entry running `agentic-plugins build` closes it: commitizen runs the hook after `version_files` are rewritten and before the bump commit, and commits with `git commit -a`, so the regenerated trees land in that same commit. `hooks.run` raises `RunHookError` on a non-zero exit, so a failed regeneration aborts the bump before any commit or tag exists.

**`agentic-plugins check` cannot catch this class, so a test was added.** It compares generated output against `_base/`, which catches a generated copy disagreeing with its source but passes when all eighteen copies agree on a stale tag. That is precisely the state the repository was in. `tests/unit/test_agent_plugin_ash_version.py` mirrors the existing `tests/unit/test_homebrew_formula_version.py`, which exists for the identical reason, and asserts three things: that each `_base/` pin matches `pyproject.toml`'s version, that each file is still listed under `version_files` so the next release maintains it, and that the `pre_bump_hooks` regeneration is configured.

Every assertion was confirmed to fail before being trusted. Three mutations, three targeted failures, nine tests running and none skipped:

| Mutation | Test that failed |
| --- | --- |
| Restore the `v3.4.0` literal in `_base/manifest.json` | `test_base_file_pins_the_packaged_version[manifest.json]` |
| Remove the `_base/mcp.json` entry from `version_files` | `test_commitizen_bumps_the_base_file[mcp.json]` |
| Remove `pre_bump_hooks` | `test_a_bump_regenerates_the_plugin_trees` |

The first mutation reproduces the original bug, and the message is the one a maintainer would want: `_base/manifest.json pins ASH v3.4.0 but this repository is version 3.7.0. Agents and users following this file would install the wrong ASH.`

## Verification

- `agentic-plugins check`: exit 0, `AGENTS.md + 17 platform outputs match _base/ source`.
- Fresh `agentic-plugins build` on the committed tree: **0 files changed**, so the committed output is exactly what `_base/` generates.
- `actionlint` 1.7.12 on the edited workflow: exit 0, zero bytes of output.
- Transpiler suite: **26 passed, 0 skipped** before and after.
- Root `tests/unit/`: **6064 -> 6073** passed, the +9 being the new file. Skips **identical at 25** on both sides.
- **No test was weakened, skipped, or deleted.**
- `ruff check` and `ruff format --check` clean. All 16 touched JSON, YAML and TOML files parse.
- `3.4.0` occurrences went from 40 to 8, with **zero** remaining in any generated file, in `_base/`, or inside the binary `ash.mcpb` bundle. The 8 that remain are not install pins: the CHANGELOG's historical `## v3.4.0` heading, a sample report `tool_version` value, an example inside a code comment, two `--context ash_version=` documentation examples whose real default is `latest`, and three lines of explanatory prose added by this PR.

Per-file diffs are one line each, except `gemini/install.sh` which carried the tag twice and `pyproject.toml` which gains the comment and the two config entries.

## Follow-ups, deliberately not fixed here

1. **`cz bump` computes 4.0.0.** The accumulated breaking change means the next bump is a major. Someone should decide whether that is intended before the next release runs.
2. **`pyproject.toml` lags the published tag.** It says `3.7.0` while v3.7.1 is the published Latest and v3.7.0 is still a Draft. Reconciling it requires moving `[tool.commitizen] version` and the four new `version_files` targets in the same change, or the silent-staleness bug returns.
3. **The other 64 tag-pinned `uses:` references**, plus an enforcement gate that fails on any unpinned reference.


rafaelpereyra (approved):
@awsmadi with all the recent changes I'm inclined to cut a new major release but let's talk about it

Issue #189, opened 2025-09-08
GitHub Actions junit reporter fails to publish results


cmckni3:
```
Error: ❌ Failed to create checks using the provided token. (TypeError: Cannot read properties of undefined (reading 'id'))
Warning: ⚠️ This usually indicates insufficient permissions. More details: https://github.com/mikepenz/action-junit-report/issues/23
```

See [GitHub Action run](https://github.com/awslabs/automated-security-helper/actions/runs/17276587645/job/49034694302). Expand junit reporter step and expand `publish results`.

awsmadi:
Resolved in PR #283 (v3.4.1 refactor/codebase-cleanup). Merged to main.

awsmadi:
Resolved in PR #283 (v3.4.1).

PR #543, opened 2026-09-09
ci: pin every remaining action to a commit sha, and gate it

awsmadi:
## What this does

Pins every remaining `uses:` reference in `.github/workflows` and `.github/actions` to a commit sha, and adds a check that fails when one drifts back to a tag.

A tag is a mutable pointer. `uses: actions/checkout@v7` resolves at run time to whatever `v7` points at then, so an upstream force-push — or a compromised maintainer account — changes the code executing inside a workflow that already holds this repository's tokens, with no diff here for anyone to review. A commit sha removes the indirection.

Three commits: the pins, the gate, and one line in the coverage-exclusions manifest so the new script is accounted for by the census that already exists.

## The count, and the count that was claimed

The working figure going in was 64 unpinned refs outside the excluded file. That is wrong. The real number is **54**.

Both a line scan and a structural YAML walk agree on the denominator, and they agree with each other:

| | count |
|---|---|
| total `uses:` in `.github/workflows/**` + `.github/actions/**` | 70 |
| already pinned to a sha | **0** |
| local (`./…`) — cannot be pinned, exempt | 14 |
| `docker://…` | 0 |
| reusable workflows called from another repo | 0 |
| unpinned (tag or branch) | 56 |
| &nbsp;&nbsp;…of those, in the excluded drift workflow | 2 |
| **unpinned and in scope here** | **54** |

Two things worth flagging about that table. Nothing in the tree was pinned before this change, so there was no existing comment style to match — the style here (`@<sha> # vX.Y.Z`) is the one used by the in-flight change described below, so the tree is consistent once both land. And the only reusable-workflow reference (`./.github/workflows/run-ash-security-scan.yml`) is local, so it needs no sha; there is also a commented-out example of the cross-repo form at `ash-repo-scan.yml:27`, which is documentation and is correctly left alone.

There are no `docker://` refs, so none were pinned. The gate does require a digest on any that appear later, so adding one does not walk through a gap in a check whose whole subject is mutable references.

## Annotated versus lightweight, and why it matters

A tag is either lightweight, where the ref already names a commit, or annotated, where the ref names a **tag object** whose sha is not a commit at all. Pinning an annotated tag's object sha produces a ref no runner can check out, and it looks correctly pinned while doing it — 40 hex characters either way, so length proves nothing.

Every ref was therefore dereferenced explicitly: `repos/OWNER/REPO/git/ref/tags/TAG`, and then `repos/OWNER/REPO/git/tags/SHA` whenever the object type came back as `tag`. Each resolved sha was then read back with `repos/OWNER/REPO/commits/SHA` to confirm it is a **reachable commit** rather than merely well-formed.

Of the 15 versions pinned, **1 required dereferencing** — `github/codeql-action` v4.38.0, whose tag object `4bd7200e…` is not the commit `b96794f0…`.

The more interesting result is in the *floating* major tags this tree was using. Three of them are annotated:

| ref in the tree | tag object (wrong) | commit (correct) |
|---|---|---|
| `astral-sh/setup-uv@v7` | `94527f2e…` | `37802adc…` |
| `github/codeql-action@v4` | `311b4ee1…` | `b96794f0…` |
| `mshick/add-pr-comment@v3` | `f4a1cfec…` | `ec328af6…` |

Pinning what those refs point at directly would have broken all three.

As a final check the shas were re-read **out of the committed files** rather than out of the resolution table, and each was re-verified upstream — so a transcription slip between resolving and writing would have surfaced. All 15 shas are reachable commits, and all 15 version comments resolve to the sha beside them.

## Versions chosen

Each ref resolves to the newest release in the major line it already used. A pin is not an upgrade, so no major boundary is crossed.

| action | was | now | commit |
|---|---|---|---|
| `actions/cache` | v6 | v6.1.0 | `55cc8345863c7cc4c66a329aec7e433d2d1c52a9` |
| `actions/checkout` | v7 | v7.0.1 | `3d3c42e5aac5ba805825da76410c181273ba90b1` |
| `actions/create-github-app-token` | v3 | v3.2.0 | `bcd2ba49218906704ab6c1aa796996da409d3eb1` |
| `actions/dependency-review-action` | v5 | v5.0.0 | `a1d282b36b6f3519aa1f3fc636f609c47dddb294` |
| `actions/github-script` | v8 | v8.0.0 | `ed597411d8f924073f98dfc5c65a23a2325f34cd` |
| `actions/setup-python` | v6 | v6.3.0 | `ece7cb06caefa5fff74198d8649806c4678c61a1` |
| `actions/setup-python` | v7 | v7.0.0 | `5fda3b95a4ea91299a34e894583c3862153e4b97` |
| `actions/upload-artifact` | v7 | v7.0.1 | `043fb46d1a93c77aae656e7c1c64a875d1fc6a0a` |
| `astral-sh/setup-uv` | v7 | v7.6.0 | `37802adc94f370d6bfd71619e3f0bf239e1f3b78` |
| `cachix/install-nix-action` | v31.11.1 | v31.11.1 | `13d8dd58da0234aa297dedd986986ccb8e7f3e24` |
| `docker/setup-buildx-action` | v4 | v4.3.0 | `37fe631027851001ddb9b187196cc803df7f5f0e` |
| `github/codeql-action/upload-sarif` | v4 | v4.38.0 | `b96794f015dfd88f77b49b1c93e0fa7110f94c63` |
| `irongut/CodeCoverageSummary` | v1.3.0 | v1.3.0 | `51cc3a756ddcd398d447c044c02cb6aa83fdae95` |
| `mikepenz/action-junit-report` | v6 | v6.5.0 | `a9170d5795813c01ab4901ffb045b52bab4ab09d` |
| `mshick/add-pr-comment` | v3 | v3.12.0 | `ec328af66588ab8f77cdeb2c264f14aba45bbf59` |

Every floating major tag already pointed at the newest patch in its line, so no ref moved to newer code than it was resolving to yesterday. Three notes:

- **`actions/setup-python` appears at both v6 and v7.** Both stay in their own line, so the same action is pinned at two different shas. That pre-existing inconsistency is left as it is rather than quietly normalized by a change whose subject says "pin".
- **`actions/github-script@v8`** has no release tagged `v8.0.0` — the release for that line is titled `v8`. The `v8` and `v8.0.0` tags point at the same commit, and `v9.0.0` exists but crossing to it would be an upgrade, so the v8 line is kept and the comment names the semver tag.
- **`actions/dependency-review-action@v5` was not a tag at all.** `v5` is a *branch* in that repository, which is a weaker guarantee than a tag. It is now the v5.0.0 commit.

## The excluded drift workflow

`.github/workflows/ash-agent-plugins-drift.yml` is untouched here. Its two refs are pinned in a separate in-flight pull request, and pinning them in both places would put two changes in conflict over the same two lines.

That leaves the gate needing to pass with those two refs still unpinned on `main`. Of the two available options — exempt the one path, or land the gate warn-only — this takes **the narrow path exemption**, because a warn-only gate is one that cannot fail, which is indistinguishable from a gate that checks nothing, and flipping it back to blocking later needs a second change that nobody may make. The exemption is narrow by comparison: it protects exactly one named file instead of weakening the rule for all 71 refs, and it is visible in the diff.

The part that makes it safe is that **the exemption expires by failing**. Following the same discipline as `typescript-coverage-exclusions.json`, where a `kind` selects a staleness test, an `--allow-unpinned` entry naming a file with no unpinned refs left is reported as stale:

```
Action pinning check failed:
  - --allow-unpinned .github/workflows/dependency-review.yml is stale: every
    reference in that file is now pinned, so the exemption excuses nothing.
    Remove the flag from the workflow that passes it.
```

So the moment the other change lands, this gate goes red and stays red until one line is deleted from `ash-actions-pinned.yml`. Forgetting the carve-out is not a quiet outcome. An exemption naming a file that has been renamed or deleted fails for the same reason.

## The gate, and proof that it fails

`.github/scripts/assert-actions-pinned.mjs` is built to the shape of the two `assert-coverage-*.mjs` scripts beside it: same argument parsing that refuses to let a flag swallow the next flag, same `problems` array written to stderr, exit 1 for a finding and 2 for a usage error, same summary line on success. It is wired into `.github/workflows/ash-actions-pinned.yml` as a checkout plus one `node` invocation.

It exempts local `./` refs (all 14 of them) because they resolve inside this repository at the commit already running, and GitHub rejects a sha on a local ref anyway. It accepts — and in fact requires — the trailing version comment.

**Requiring the comment is not a style rule.** `.github/dependabot.yml` declares the `github-actions` ecosystem for `/` on a daily schedule, and Dependabot reads `# v7.0.1` to learn which version a sha represents. A pinned-but-uncommented ref quietly removes itself from the upgrade path: nothing goes visibly stale, updates simply stop being offered.

### Mutation proof

A gate that cannot be shown failing is indistinguishable from a gate that matches nothing. Reverting one pinned ref to a tag and stripping one version comment:

```
$ node .github/scripts/assert-actions-pinned.mjs \
    --allow-unpinned .github/workflows/ash-agent-plugins-drift.yml
Action pinning check failed:
  - .github/workflows/dependency-review.yml:28 actions/checkout@v7 is pinned to
    "v7", which is a tag or branch, not a commit sha. […]
  - .github/workflows/dependency-review.yml:29
    actions/dependency-review-action@a1d282b36b6f3519aa1f3fc636f609c47dddb294 is
    pinned to a sha but carries no trailing version comment. […]
EXIT=1
```

Both mutations are caught, each naming its file and line. Restoring them:

```
$ node .github/scripts/assert-actions-pinned.mjs \
    --allow-unpinned .github/workflows/ash-agent-plugins-drift.yml
action pinning OK: 71 `uses:` reference(s) across 20 file(s) -- 55 pinned to a
commit sha, 14 local and exempt by nature, 2 unpinned under an explicit
--allow-unpinned exemption that is still needed
EXIT=0
```

The gate checks its own workflow file too, which is why the total is 71 rather than 70.

Other failure paths, each exercised: unknown argument and a flag with no value exit 2; an exemption naming a file that is not in the census exits 1; a stale exemption exits 1.

### No vacuous passes

The realistic way a check like this fails is by matching nothing and exiting 0, so each route to an empty examination is closed: an empty census fails, finding zero refs fails, finding refs but none a sha could apply to fails, and any line that begins with `uses:` but whose value the scanner cannot read is reported rather than skipped.

That last one is also why the scan is anchored to the first token on the line instead of searching for `uses:` anywhere in it. **`statuses: write` and `if not statuses:` both contain the substring `uses:`, and both appear in this tree** — a loose pattern reports both as unpinned refs, and a gate that cries wolf on correct configuration does not survive contact with a reviewer.

What the check cannot do is tell a real commit sha from a 40-character tag name; that needs a network call. It guards against a tag being *reintroduced*, and the API verification above is the other half of the guarantee.

### Triggers

`pull_request` and `merge_group`, with **no `paths` filter on either**. Six workflows here already run on `merge_group`, so a gate without it would never report for a queued entry and would block the merge queue on a verdict that never arrives.

`merge_group` accepts no `paths` filter, which was verified rather than assumed:

```
$ actionlint mg-paths.yml
mg-paths.yml:7:5: "paths" filter is not available for merge_group event.
  it is only for push, pull_request, pull_request_target events [events]
$ actionlint mg-badtype.yml
mg-badtype.yml:4:13: invalid activity type "opened" for "merge_group" Webhook
  event. available types are "checks_requested" [events]
```

Filtering `pull_request` alone would leave the gate running in one event and not the other. Since the check censuses the whole tracked tree rather than the diff, it is meaningful on every run regardless of what changed, and the job installs no dependencies — so there was nothing to save by filtering, and no asymmetry to reconcile.

**This adds a new check name.** It only blocks once someone adds `Assert every action is pinned to a commit sha` to the required checks for `main` and for the merge queue; that is a repository setting and not something this pull request can do.

## Verification

- **`actionlint` 1.7.7**, exit 1 on `.github/workflows/*.yml` — but with **byte-identical output to the `origin/main` baseline**: 9 pre-existing findings (2 empty `needs:` sections in `ash-repo-docs.yml`, 7 informational shellcheck SC2086 in `run-ash-security-scan.yml`), **0 introduced**. On the new workflow alone, and on the 9 changed workflows that carry no pre-existing findings, actionlint exits 0. The four changed composite `action.yml` files produce output only because actionlint parses any file handed to it as a *workflow* and has no composite-action mode; that output is also identical before and after.
- **Every changed YAML parses** — 19 files before, 20 after, no parse failures. The line scan and the structural walk agree on 70 refs before and 71 after.
- **No action name changed.** Checked mechanically rather than by eye: all 70 `uses:` lines were paired by file and line number between `origin/main` and this branch, and for each pair the path before `@` was compared byte for byte. 54 refs changed tag→sha, 16 untouched (14 local + 2 in the excluded file), **0 names changed**. The diff is 54 insertions and 54 deletions across 14 files — one line each, nothing added or removed. `github/codeql-action/upload-sarif` keeps its subdirectory path.
- **Tests that read workflow files: 211 passed, 0 failed, 0 skipped.** Six test files reference workflow or action files changed here, and all were run on Python 3.13 with the project environment: `test_release_workflow_token.py`, `test_junit_permission_check.py`, `test_external_target_scan_gate.py`, `test_multi_project_attribution_gate.py`, `tests/unit/utils/test_ignore_paths_glob_matching.py`, `tests/unit/utils/test_sarif_suppressions_extended.py`. One of those six, `test_junit_permission_check.py`, has since been rewritten -- see *Follow-up* below for why deleting it was not an option. The other five are unmodified. Two of these were at real risk: `test_the_token_action_is_pinned_to_a_major_version` asserts `actions/create-github-app-token@` is present and `@main` is absent, which a sha satisfies more strictly than a tag; and `test_checkout_does_not_receive_the_app_token` locates its step by matching `actions/checkout` inside the `uses:` string, which is why keeping the action name byte-identical mattered.
- **`pre-commit run --files` on the changed files**: `generate-schemas` passed, ASH's own scan passed, `ruff` and `ruff-format` reported no files to check at that point. One Python file has since been added under `.github/` and one test file rewritten; both are `ruff check` and `ruff format --check` clean.

## One pre-existing problem found, deliberately not fixed here

`pretty-format-json --autofix` reformats `typescript-coverage-exclusions.json` by sorting keys within each entry, and it rewrites **all four** entries — so the file on `main` already does not satisfy that hook. Nothing in CI runs `pre-commit` or `ruff` over this repository's own code, so this is a local-developer hook rather than a gate. The autofix was reverted so this change does not reorder three unrelated entries as a formatting-only edit bundled into a behavioral one; the new entry matches the key order of its three siblings.

Separately, and more worth someone's attention: **`assert-coverage-completeness.mjs --scope outside` already exits 1 on `main`**, with four findings that predate this change.

```
Coverage completeness check failed:
  - .github/scripts/assert-coverage-completeness.mjs is measured by no coverage
    report and is named in no exclusion. […]
  - .github/scripts/assert-coverage-scope.mjs is measured by no coverage report
    and is named in no exclusion. […]
  - exclusion deploy/cdk/jest.config.js names a file that is not tracked
    (deleted or renamed). Remove the entry.
  - exclusion deploy/cdk-constructs/jest.config.js names a file that is not
    tracked (deleted or renamed). Remove the entry.
```

Both assert scripts are tracked production `.mjs` outside either package scope and named in no exclusion, and both `deploy/*/jest.config.js` entries are stale because `deploy/` does not exist on the default branch yet. The job is dormant rather than red: `ash-typescript-ci.yml` is path-filtered to `deploy/cdk/**` and `deploy/cdk-constructs/**`, so with `deploy/` absent nothing triggers it.

Fixing those four is a change to the TypeScript coverage gate and belongs with whatever lands `deploy/`, not in a change about action pinning. What the one-line manifest addition here buys is that the count does not grow to five on the way past — after this change the new script is accounted for and the remaining findings are exactly the four above, unchanged.

## Follow-up: three review findings, fixed in four commits on top

An adversarial review of this branch found three problems. All three are fixed
without rewriting any published history -- the four commits below sit on top of
`7af60d4`.

### 1. Five tests asserted the step this branch deleted (was blocking)

`TestJunitPermissionPreCheck` in `tests/unit/test_junit_permission_check.py`
required the `Verify checks write permission` probe, its `id: check-perms`, and the
`if:` that referenced it. This branch removed all of that, so those five failed on
all 18 `unit-test` matrix cells and took `required-checks` down with them.

They were not simply deleted. That file is the only regression coverage for #189,
and the probe it pinned was a **misdiagnosed fix**: it called
`GET /repos/{owner}/{repo}/check-runs`, a route that does not exist, so it 404'd for
everyone, `has_permission` was always false, and the publish step never ran once. A
GET cannot establish WRITE access in any case. The real cause was `update_check:
true` selecting the action's update path, which resolves a check run by `job_name`
(default `github.job`, a YAML key) while GitHub names check runs by display name --
so the lookup matched nothing, `annotator.ts` read `check_runs[0].id` unguarded, and
the step threw, swallowed the error, reported success and published no annotations.

The five assertions are replaced by three that pin the actual cause:

- **No junit-report step anywhere under `.github/` sets `update_check` truthy.** Swept
  across every call site rather than hardcoded to the two that exist, so a third call
  site is covered too. Carries a positive-control test that fails if the sweep ever
  matches nothing.
- **The scan workflow's `report_paths` is not a `**` glob**, because `**` does not
  descend into hidden directories and the report lands under `.ash/ash_output/`.
  Scoped to this workflow: `run-unit-tests` globs `**/pytest.junit.xml` correctly,
  since `test-results/` is not hidden.
- **The fork condition gates the publish step**, which is the condition the deleted
  probe was reaching for.

Each was mutation-tested against the pre-fix state. Restoring `update_check: true`
fails assertion 1 only; making `report_paths` a `**` glob fails assertion 2 (both of
its tests); stripping the fork condition fails assertion 3; renaming the action so
the sweep under-matches fails the positive control and names the surviving call site.
The class docstring and the file header now explain that the original diagnosis was
wrong and what replaced it.

### 2. BEHAVIOR CHANGE: downstream consumers get a red check where they had none

`run-ash-security-scan.yml` is `on: workflow_call` with `collect-junit-xml-report`
defaulting to `true`. Because the probe was always false, **no
`ASH Scan JUnit Test Report` check run was ever created for any caller.** Removing
the gate makes the step run for the first time.

The action derives the check run's conclusion from the JUnit failure count, and ASH
encodes **actionable findings as test failures**. So any downstream repository
calling this reusable workflow with a single actionable finding goes from "no check
at all" to a **red check**, with no diff of its own.

Measured on this pull request: check run `ASH Scan JUnit Test Report` = **failure**,
`216 tests run, 0 passed, 210 skipped, 6 failed`. Control: no such check run exists
on #559 or #560, which do not carry the fix.

**The fix is kept and the documentation is made true**, which is the same call taken
on the analogous case in #559. The default is deliberately *not* flipped:
`fail_on_failure` stays at its default of `false`, so the step and the job both stay
green and no build fails because of this. The residual cost is a red check in the
checks list that counts toward the commit's combined status.

The one case that would have justified flipping the default instead is a branch
protection rule requiring this check by name. That cannot be inherited: the check run
never existed before, so no rule could already require it.

The warning is written out at the `collect-junit-xml-report` input declaration, with
a short form in the input `description` so it renders wherever the `workflow_call`
inputs are documented, and a pointer from the publish step -- which is what someone
debugging an unexpected red check will be reading. It is **not** in `CHANGELOG.md`:
`pyproject.toml` sets `changelog_file = "CHANGELOG.md"` with
`update_changelog_on_bump = true`, so commitizen regenerates that file from commit
messages on version bump and a hand-written entry would be clobbered.

Callers who want the previous silence can set `collect-junit-xml-report: false`; the
results stay in the job summary and the `ash_output` artifact.

### 3. The `grep -c` change repaired a failure that never happened

The comment added at three sites claimed the unguarded `grep -c` "failed this step in
the good case". **It did not, and the repository's own CI shows it.** `TextReporter`
emits `"  - ERROR = Scanner execution error"` whenever `include_summary` is true,
which is its default and is set explicitly in both configs CI uses, so the grep always
matched at least once and always exited 0. Run `34493384101` at head `518cc13` still
carried the unguarded form with no `|| true` at all three sites, and the
`scan (bash, ubuntu-latest)`, `scan (python-local, ubuntu-latest)` and
`scan (python-container, ubuntu-latest)` cells all reported **success**. The `|| true`
was a no-op.

The real defect is the **direction** of the failure. The `-gt 1` threshold exists only
to skip that legend line. With the legend present, one real scanner `ERROR` counts 2
and trips the gate; if the legend wording ever changes or `include_summary` is turned
off, one real `ERROR` counts 1, `-gt 1` is false, and the step passes. **Exactly one
scanner error reads as clean.** A loud wrong answer was on its way to becoming a
silent one.

While fixing that, a fourth site turned out to be worse and already broken. The
PowerShell leg iterated `$Results.scanners.<name>.result`. `AshAggregatedResults`
declares **`scanner_results`**, and each entry carries **`status`**, not `result`. The
property set was empty on every run, `$ErrorCount` stayed 0, and the step printed
`SUCCESS: No plugin execution errors found` **without ever inspecting a scanner** --
and it was the implementation the other legs were about to be modelled on. The Windows
`python-local` leg also used `Select-String "ERROR"`, which is case-insensitive by
default, so any report line containing lower-case "error" inflated its count on top of
the legend offset.

All five now call one shared script,
`.github/actions/run-scan-test/count_scanner_errors.py`, which reads `scanner_results`
from `ash_aggregated_results.json` and counts entries at status `ERROR` -- so no legend
line exists to skip and the threshold is a plain `> 0`. It normalizes the status the
way `scripts/verify_external_target_scan.py` does, so `"ERROR"`, `"error"` and
`"ScannerStatus.ERROR"` all count; it inspects both the flat `ScannerTargetStatusInfo`
shape written by `unified_metrics` and the nested `source`/`converted`
`ScannerStatusInfo` shape other paths assign; and it **exits non-zero when the file is
missing, unparseable, or has no `scanner_results` or an empty one**, rather than
reporting zero errors from an empty input.

`setup-ash` runs `actions/setup-python` for every `scan-validation` cell, including the
`bash` and `powershell` methods, so one Python implementation covers all five legs. The
two `pwsh` sites resolve `python3`-or-`python` rather than assuming either.

### Verification of the follow-up

- **Suite: 6 failed -> 1 failed.** Before, at `7af60d4`:
  `6 failed, 6566 passed, 119 skipped, 3 xfailed` (6694 total). After:
  `1 failed, 6572 passed, 119 skipped, 3 xfailed` (6695 total -- the rewrite is 10 tests
  where there were 9). The single remaining failure is
  `test_agent_plugin_ash_version.py::TestInstallRefsTrackPackagedVersion::test_every_install_ref_names_the_packaged_version`,
  which is **inherited from `origin/main` and fixed by #558**, not by this branch.
- **Each replacement assertion was proved able to fail**, by four separate mutations of
  the workflow; see the list in section 1.
- **The ERROR counter was probed at 0, 1 and 2**, returning `0`, `1` and `2` and exiting
  0 / non-zero / non-zero. One `ERROR` is now detected -- the off-by-one the old `-gt 1`
  threshold missed. A missing or empty `scanner_results` exits non-zero rather than
  reporting 0, and a scanner *named* `error-prone-scanner` at `PASSED` counts 0.
  12 of 12 probe cases pass.
- **The pinning gate is unmoved**, as predicted before running it, because no `uses:`
  line was added or removed:
  `85 uses reference(s) across 21 file(s) -- 70 pinned to a commit sha, 15 local and
  exempt by nature, 0 unpinned`, exit 0.
- **`assert-coverage-completeness.mjs --scope outside` still passes** -- the new file is
  Python and that census covers TypeScript/JS only.
- **Every touched YAML parses.** `actionlint` was **not available** in the environment
  the follow-up was verified in, so the workflow check here is a YAML parse of both
  touched files rather than a lint. The bash step bodies were parse-checked under
  `ERR_EXIT`/`PIPE_FAIL`. **`pwsh` was also unavailable**, so the two PowerShell
  snippets were reviewed but not executed.
- **Scope**: only `.github/` and `tests/unit/test_junit_permission_check.py` are
  touched. No change to `.ash/.ash.yaml` and no Python outside that one test file and
  the new counter under `.github/`.

### A fourth problem the follow-up caused, and fixed

Worth recording because it is a failure mode this repository is unusually exposed
to: **the replacement test's own failure message tripped ASH's self-scan.**

The `update_check` assertion's message concatenated `"...selects the action's
update path..."` and `f"Set by: {offenders}"`. Read as one string that is
`UPDATE ... SET <variable>`, which is exactly bandit's `B608`
(`hardcoded_sql_expressions`) heuristic for string-built SQL. bandit reported it
**MEDIUM**, MEDIUM is actionable at this repository's own threshold, and so
`Validate ASH using Python Local (Unix)` exited 2 with `1 actionable finding`.
That reddened `scan (python-local, ubuntu-24.04-arm)`, its Community Plugins
variant, `scan (python-container, ubuntu-24.04-arm)` and
`podman (ubuntu-24.04-arm, py3.12)` -- four cells that were green at `7af60d4`.

Fixed by rewording the prose so the two keywords no longer straddle the
interpolated value. **No `nosec`**: suppressing a live bandit rule so a sentence
could keep its phrasing trades a scanner rule for wording.

One measurement note that mattered here. Running `bandit` on the new counter at
`.github/actions/run-scan-test/count_scanner_errors.py` reported *zero findings* --
with `loc: 0`. bandit had skipped the file for being under a dot-directory, so
"zero findings" was not a claim about the file at all. Re-run against a copy at a
non-hidden path it reports `loc: 199` and zero findings at every severity, which is
the actual result. The test file measures `loc: 178` and `SEVERITY.MEDIUM 0` after
the reword, against `MEDIUM 1` before.

### Measured CI outcome at `5d8d605`

- **All 24 `scan (...)` cells green**, including all five legs that now call the
  shared counter. Verified that the counter *ran* rather than being skipped: the
  Windows `python-local`, Linux `bash`, Linux `powershell` and Linux
  `python-container` logs each print `No scanner reported status ERROR.` That line
  is only reachable after a non-empty `scanner_results` was successfully read --
  a missing file or an empty mapping exits 1 -- so it is evidence of inspection,
  not of silence. For the `powershell` leg this is the first time the step has ever
  inspected a scanner.
- **`Assert every action is pinned to a commit sha`: success**, reporting
  `85 uses: reference(s) across 21 file(s) -- 70 pinned to a commit sha, 15 local
  and exempt by nature, 0 unpinned`. Unchanged, as predicted, because no `uses:`
  line was added or removed.
- **37 checks still red, and every one of them is accounted for:** 18
  `unit-test (...)` cells, the 18 `PyTest - ...` check runs those cells publish, and
  `ASH Scan JUnit Test Report`.
  - The 18 `unit-test` cells all fail on the single inherited test
    `test_every_install_ref_names_the_packaged_version` ("These install references
    do not name the packaged version 3.7.0"), which is red on `origin/main` and is
    fixed by #558, not here. Sample counts: `ubuntu-24.04-arm py3.12` reports
    `6695 tests run, 6572 passed, 122 skipped, 1 failed`; `windows-latest py3.13`
    reports `6695 tests run, 6501 passed, 193 skipped, 1 failed`. Both are **1
    failed**, down from the 6 this branch had before the follow-up.
  - `ASH Scan JUnit Test Report` is red at `216 tests run, 0 passed, 210 skipped,
    6 failed` -- **this is the documented behavior change from section 2 above,
    happening on this very pull request.** It is the intended consequence of the
    junit fix working, and the reason the input declaration now carries a warning.
- Ten macOS `install-methods` cells and `iac-drift` were **still queued** when this
  was written, so they are reported as pending rather than passing.


rafaelpereyra (dismissed):

nikmeiser (approved):
CI green except the one known red — "ASH Scan JUnit Test Report" (a
0-second report-publisher/annotation-permission failure); all real scans/gates pass, including "Assert every action
is pinned to a commit sha" (7s), CodeQL, dependency-review, the full PyTest matrix, and "ash / SAST, SCA, and IaC
Scan".

PR #551, opened 2026-09-09
fix(container): compare the OCI runner's basename in the GHA layer-cache gate

awsmadi:
## The defect

`_gha_layer_cache_args` compared an already-resolved runner path against the bare string `"docker"`, so the comparison failed for every runner and the `type=gha` layer cache had never been enabled for anyone.

Repairing that comparison made the path execute for the first time, which exposed three further problems in the code it turned on. All four are fixed here.

## The verified trace, and what each step returns

| Step | Returns |
| --- | --- |
| `_resolve_oci_runner(oci_runner)` (`run_ash_container.py:277`) | the return value of `_find_runner`, not the candidate name — `return path` |
| `_find_runner(name)` | `subprocess_utils.find_executable(name)` verbatim |
| `find_executable(command)` (`utils/subprocess_utils.py:26`) | an absolute path on both branches: `shutil.which(cmd)`, or `poss.as_posix()` for the `ASH_BIN_PATH` and `/usr/local/bin` fallbacks. `None` only when nothing is found, and `_resolve_oci_runner` then raises `RuntimeError` |
| `_gha_layer_cache_args(resolved_oci_runner, ...)` | compared that absolute path against `"docker"` |

`resolved_oci_runner` is assigned once and passed straight through to `_build_image` and on to the gate. **There is no normalization step anywhere between them.** This holds even when the runner is given explicitly: `OCI_RUNNER=docker` still goes through `find_executable`, so the gate still sees a path.

Measured on a host where `docker` is `/usr/bin/docker`, with the Actions cache environment set so every other guard passes:

```
find_executable('docker')                        -> '/usr/bin/docker'
_gha_layer_cache_args('/usr/bin/docker',    ...) -> cache declined -> []
_gha_layer_cache_args('docker',             ...) -> CACHE ENABLED
```

Only the bare name — the one shape production never produces — was accepted.

## Why this survived: the tests passed an input no caller supplies

`test_gha_layer_cache_args.py` did assert a non-empty return, and `test_run_ash_container_helpers_coverage.py` even asserted the full `docker buildx build --load` argv end to end. Both passed the literal `"docker"`. So the positive case was covered with a value the resolver cannot produce, and the suite stayed green while the only value the gate ever receives was rejected. Coverage was not the gap; the input shape was.

The fix is `Path(resolved_oci_runner).stem != "docker"`, which accepts `/usr/bin/docker`, a bare `docker`, and the `.exe` form `find_executable` builds on Windows, while declining `/usr/bin/docker-compose` and the other `docker`-prefixed binaries that share a directory with it. `_resolve_oci_runner` is called exactly once and no other bare-name comparison against a resolved runner exists in `automated_security_helper/`, so this is one instance rather than a class.

## What turning the path on actually revealed

The first push of this branch failed three cells that a control PR passes: `scan (python-container, ubuntu-latest)`, `scan (python-container, ubuntu-24.04-arm)`, and `scan (python-container, ubuntu-latest) [offline]`. That set is exactly `method: python-container` and `oci-runner: docker` and `contains(os, 'ubuntu')` — precisely the condition under which the repaired gate returns flags. Nothing else moved. (The red `unit-test` cells are a `mikepenz/action-junit-report` token-permission error that appears identically on the control PR; the tests themselves passed and uploaded results.)

### 1. `ignore-error=true` on `cache-to`, and only on `cache-to`

The configuration was correct — buildx on the `docker-container` driver, credentials present, argv well-formed. The cache service returned an HTML error page:

```
23:54:06  #6  importing cache manifest from gha:10602703455898056127
23:54:06  #6  ERROR: failed to parse error response 400: <h2>Our services aren't available right now</h2>...
23:54:17  #13 [core 7/28] RUN apt-get update && ...      <- build continued, cache miss
23:56:48  #46 exporting to GitHub Actions Cache
23:56:48  #46 ERROR: failed to parse error response 400: <h2>Our services aren't available right now</h2>...
23:56:49  ERROR: failed to build: failed to solve: ...
```

The import failure at `#6` was non-fatal — the build ran on for another 2m42s. The export failure at `#46` was fatal, and the propagated error carries `#46`'s edge-reference id rather than `#6`'s. So buildkit already degrades gracefully on `cache-from` and dies on `cache-to`, and only the write direction needs wrapping. `cache-from` is deliberately left bare so real read problems stay visible.

An earlier revision of this description argued that `ignore-error` would mask misconfiguration and left it out. That reasoning does not survive this run: there was no misconfiguration to mask, only a transient upstream outage taking a security scan down with it. The claim has been withdrawn.

**The fix is confirmed by the same failure recurring and being absorbed.** The cache service returned the identical 400 on the export step of the re-run, and the cell passed anyway:

```
01:01:55  #46 exporting to GitHub Actions Cache
01:01:55  #46 preparing build cache for export 0.0s done
01:01:57  #46 sending cache export 1.6s done
01:01:57  #46 ERROR: failed to parse error response 400: <h2>Our services aren't available right now</h2>...
          (no "failed to build: failed to solve" anywhere in the job)
scan (python-container, ubuntu-latest)   pass   6m35s
```

Same step, same error, opposite outcome: before this change that `#46` 400 produced `ERROR: failed to build: failed to solve` and exit 1. An import 400 recurred at `#4` in the same run and was tolerated as before. So the cache is genuinely live rather than silently declining — the log shows the probe running as `/usr/bin/docker buildx version`, the new scope on both directions, and both an import and an export attempt.

### 2. The gate trusted the filename, which is not the capability

podman's `podman-docker` package installs a `/usr/bin/docker` script that emulates the Docker CLI by exec'ing podman, and `_OCI_RUNNER_CANDIDATES` reaches `docker` before `podman`. On a RHEL or Fedora host with that package and no finch, the resolved runner is named `docker`, has no buildx, and passes every name-based check.

That matters more than a lost optimization: before the comparison fix the gate rejected everything and the shim worked, so a name-only check would be a hard regression for that population. ASH is a distributed public scanner, so third-party CI is a real user base for it.

No path accessor can tell these apart, so the gate now asks: `<runner> buildx version`, memoized per resolved path. It runs last, after the environment checks have already established we are inside Actions, so it shells out about once per build and never on a local run that was going to decline anyway.

### 3. The scope key pinned a constant and omitted both dimensions that vary

`scope` was `f"ash-{build_target}"`, and `build_target` cannot discriminate on any path that reaches the cache: it is forced to `"ci"` whenever `CI` is set, and the cache requires `ACTIONS_RUNTIME_TOKEN`, which exists only inside Actions, where `CI` is always set. So the scope was the constant `ash-ci` and the `ci`/`non-root` collision the old comment guarded against was unreachable.

What does vary across the three cache-enabled cells is CPU architecture and `OFFLINE`. The logs show the two architectures requesting different manifests under that one scope (`gha:10602703455898056127` on amd64, `gha:16327726473344493535` on arm64), so they demonstrably cannot share entries, and buildkit keeps only whichever build finished last. `ARG OFFLINE` is consumed at `Dockerfile:53` and exported as `ENV` at line 61, so an offline and an online build share almost no layers below that point. Both are now in the key.

**`mode=max` is dropped in favor of `mode=min`, and this is the reason.** Tripling the scope count while exporting every intermediate stage of a multi-GB scanner image runs straight at a documented 10 GB per-repository cache cap that evicts by least-recent access **across all workflows in the repository**. The most likely casualty is the `astral-sh/setup-uv` cache that the 18-cell unit-test matrix restores — a regression that would surface in a different job and be nearly impossible to attribute. `mode=min` bounds the export to the layers of the target image, which is what a rebuild of the same cell reuses anyway.

A shared fallback scope on `cache-from` was considered and rejected: nothing writes such a scope, so reading it would be dead weight.

## Both-directions test proof

Against the original single-scope, `mode=max`, no-`ignore-error` code, 12 of the new assertions fail, including `test_only_cache_to_ignores_errors`, `test_export_is_mode_min`, `test_architecture_is_in_the_scope`, `test_offline_is_in_the_scope` and `test_both_directions_use_the_same_scope`. Against the original bare-string comparison, the four resolved-docker-path cases fail. All pass after.

The Windows case is asserted on the flavor that produces it. A backslash path only resolves correctly under `WindowsPath`; under a POSIX interpreter `PurePosixPath(r"C:\...\docker.exe").stem` is the entire string. Production is correct because `Path` dispatches to `WindowsPath` on Windows, but the previous revision's comment claimed a forward-slash spelling made the case platform-neutral, which was false and would have talked the next reader out of checking. The test now dispatches the module's `Path` to `PureWindowsPath`, and removing that dispatch makes it fail. This is the only coverage the `.exe` branch has anywhere: no Windows cell in CI can reach this gate, because the credentials are gated on ubuntu and `windows-latest` appears in the scan matrix only as `python-local`.

`test_scope_is_per_build_target` keeps its assertion but now records that `"non-root"` is a value the cache-enabled path cannot produce, with the reachable discriminators covered separately — the same trap this PR diagnoses for the runner, eight lines above the new class. The identical trap in `test_run_ash_container_helpers_coverage.py` is corrected the same way. Both files now stub the capability probe so they do not depend on whether the machine running them has docker installed.

## Verification

Full suite as CI runs it — `uv sync --extra cdk` then `uv run pytest`:

| | passed | skipped | xfailed | failed |
| `origin/main` baseline | 6561 | 119 | 3 | 0 |
| this branch | 6584 | 119 | 3 | 0 |

The +23 are the new cases. Skipped is unchanged at 119 and xfailed at 3, so nothing was skipped or turned into an expected failure.

Lint, compared against an `origin/main` baseline across all three touched files: 46 findings before, 46 after, zero added and zero removed — all pre-existing `UP006`, `DTZ005`, `BLE001` and `SIM102`. The capability probe catches `(OSError, SubprocessError)` rather than a bare `Exception` specifically so it does not add a 5th `BLE001`; that is safe because `run_command` with `check=False` already converts a timeout or a missing binary into a non-zero `CompletedProcess` and never raises. Both test files are clean under `ruff check` and `ruff format --check`. `run_ash_container.py` would be reformatted both before and after, at the same 11 hunks, none touching the changed lines.

## What was not measured

**The cache's effect on build time was not measured**, and this revision still does not measure it. What is now shown is narrower: the flags are emitted, and a cache-service failure no longer fails the build. Whether the cache shortens the container build, and by how much, is not established here.

The eviction risk above is likewise reasoned from the documented 10 GB cap and its cross-workflow LRU policy, not from an observed eviction. `mode=min` is a precaution against it, not a measurement of it.

### DIFF
--- a/automated_security_helper/interactions/run_ash_container.py
+++ b/automated_security_helper/interactions/run_ash_container.py
@@ -368,7 +368,13 @@ def _gha_layer_cache_args(
     the run's token. Distributing an image is a separate ``--push`` operation that
     this does not perform.
     """
-    if resolved_oci_runner != "docker":
+    # Compare the final path component, not the whole string. The runner arrives
+    # here already resolved: _resolve_oci_runner returns find_executable's output,
+    # which is a full path such as /usr/bin/docker, and carries a .exe suffix on
+    # Windows. ``stem`` reduces every one of those shapes -- and a bare name -- to
+    # ``docker``, while still declining neighbors like /usr/bin/docker-compose
+    # that a substring test would wrongly accept.
+    if Path(resolved_oci_runner).stem != "docker":
         return []
     if force:
         return []
--- a/automated_security_helper/interactions/run_ash_container.py
+++ b/automated_security_helper/interactions/run_ash_container.py
@@ -2,13 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import json
+import logging
 import os
+import platform
 import re
 import shlex
 import shutil
 from subprocess import (
     CalledProcessError,
-)  # nosec B404 - Using the exception class to evaluate subprocess invocations
+    SubprocessError,
+)  # nosec B404 - Using the exception classes to evaluate subprocess invocations
 import sys
 import time
 from datetime import datetime
@@ -43,6 +46,9 @@ _SAFE_REVISION_PATTERN = re.compile(r"^[A-Za-z0-9._\-/]+$")
 
 # Discovery order for OCI runners (first found wins)
 _OCI_RUNNER_CANDIDATES = ["finch", "docker", "nerdctl", "podman"]
+
+# Memoizes the buildx capability probe, keyed by resolved runner path.
+_BUILDX_SUPPORT_CACHE: dict[str, bool] = {}
 
 
 def _validate_ash_revision(revision: str) -> bool:
@@ -341,10 +347,46 @@ def _find_dockerfile(resolved_revision: str | None) -> Path:
     return dockerfile_path
 
 
+def _runner_supports_buildx(resolved_oci_runner: str) -> bool:
+    """Whether the runner actually provides ``buildx``, probed once per path.
+
+    The filename does not establish the capability. podman ships a
+    ``podman-docker`` package whose ``/usr/bin/docker`` is a shim that execs
+    podman, and ``_OCI_RUNNER_CANDIDATES`` reaches ``docker`` before ``podman``, so
+    on such a host every name-based check passes for something that has no buildx
+    at all. Asking the binary is the only thing that separates the two.
+    """
+    if resolved_oci_runner in _BUILDX_SUPPORT_CACHE:
+        return _BUILDX_SUPPORT_CACHE[resolved_oci_runner]
+
+    supported = False
+    try:
+        result = subprocess_utils.run_command(
+            [resolved_oci_runner, "buildx", "version"],
+            capture_output=True,
+            check=False,
+            log_level=logging.DEBUG,
+            timeout=30,
+        )
+        supported = result.returncode == 0
+    except (OSError, SubprocessError) as e:
+        # run_command with check=False already converts a timeout or a missing
+        # binary into a non-zero CompletedProcess, so this is belt and braces.
+        ASH_LOGGER.debug(f"buildx probe for {resolved_oci_runner} failed: {e}")
+
+    if not supported:
+        ASH_LOGGER.debug(
+            f"{resolved_oci_runner} does not provide buildx -- skipping layer cache"
+        )
+    _BUILDX_SUPPORT_CACHE[resolved_oci_runner] = supported
+    return supported
+
+
 def _gha_layer_cache_args(
     resolved_oci_runner: str,
     build_target: str,
     force: bool,
+    offline: bool,
 ) -> List[str]:
     """Return buildx flags that read and write image layers via the GitHub Actions cache.
 
@@ -355,6 +397,7 @@ def _gha_layer_cache_args(
 
     - docker only. ``type=gha`` is a buildx cache backend. podman and finch expose
       only registry-backed caches, so there is nothing equivalent to offer them.
+    - And docker in fact, not just in name. See ``_runner_supports_buildx``.
     - Only inside a GitHub Actions run. The exporter authenticates with
       ACTIONS_RUNTIME_TOKEN and talks to ACTIONS_CACHE_URL (protocol v1) or
       ACTIONS_RESULTS_URL (v2). Note that Actions does not place these in the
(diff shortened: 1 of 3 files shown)


awsmadi:
Superseded by #560, which consolidates this with #554 and #552 under one theme: operations that claim to retry or cache actually do. The layer-cache work lands there unchanged -- the `Path(...).stem` comparison, the memoized buildx probe, `ignore-error=true` on the export direction only, the arch-and-OFFLINE scope, and `mode=min` -- with the resolved-path tests shown failing against unmodified code first. Branch kept.

PR #554, opened 2026-09-10
fix(retry): make the configured backoff actually happen

awsmadi:
## What this fixes

Three retry mechanisms each announced a delay they did not take. All three are the same shape: the diagnostic message is emitted *upstream* of the arithmetic that computes the interval, so the message keeps printing after the interval is gone.

### 1. `install_tool_with_version` accepted `retry_config` and ignored it

`automated_security_helper/utils/uv_tool_runner.py` — the function took a `retry_config: Optional[UVToolRetryConfig]` parameter, documented "retry logic" in its docstring, and had **no retry loop at all**. A caller asking for `max_retries=3` got exactly one attempt.

The parameter is not decorative. There is a live chain:

- `plugin_modules/ash_builtin/scanners/bandit_scanner.py` builds `{"max_retries": 3, "base_delay": 1.0, "max_delay": 60.0}`
- `base/uv_tool_mixin.py` converts it to a `UVToolRetryConfig` and logs `[INSTALLATION_CONFIG] Using custom retry configuration / max_retries=3, base_delay=1.0s`
- then passes `retry_config=config` into a function that never read it

Measured with `subprocess.run` stubbed to fail the way a reset connection does:

| | attempts | sleeps |
|---|---|---|
| before | 1 | none |
| after | 4 | 1.71s, 2.64s, 4.61s |

`retry_config=None` still means a single attempt, so every caller that never asked for retries keeps its behavior exactly. A timeout stays terminal on purpose: that attempt already spent the whole `timeout` budget, so retrying would multiply wall clock by the attempt count, and three 300s timeouts is fifteen minutes of a scan phase waiting on a tool that is not coming.

### 2. `retry_with_backoff` could hand `time.sleep` a negative number

`plugin_modules/ash_aws_plugins/aws_utils.py` — `base_delay` and `max_delay` reach this decorator from reporter config, and `CloudWatchLogsReporter` and `S3Reporter` both declare them as bare `float` fields with no lower bound. A negative `max_delay` made the computed interval negative; `time.sleep` raises `ValueError` on a negative argument; and the broad `except Exception` in `_create_log_stream_with_retry` swallows it. The retry silently became zero retries and the warning blamed the log stream.

The interval is now clamped non-negative. The floor is `0.0` rather than some small positive number because `base_delay=0` is a legitimate way to ask for immediate retries — what must not happen is a negative interval reaching `sleep`, or an exception escaping the arithmetic that exists to protect a throttled API.

### 3. `with-retry.sh` truncated its own retry loop and misreported it

`automated_security_helper/assets/with-retry.sh` — `sleep $delay; delay=$((delay * 2))` is bash integer arithmetic, which cannot parse a value like `0.1`, the obvious way to ask for a fast-but-nonzero backoff. That arithmetic error aborts the enclosing `while` **without** aborting the script (there is no `set -e`), so exactly one attempt ran and the script still printed `All 3 attempts failed` and exited 1.

Observed directly, with a marker file counting executions:

```
$ WITH_RETRY_DELAY=0.1 bash with-retry.sh 'printf x >> marker; false'
Attempt 1/3 failed, retrying in 0.1s...
with-retry.sh: line 29: 0.1: syntax error: invalid arithmetic operator (error token is ".1")
All 3 attempts failed
$ cat marker
x          # one attempt, not three
```

Both knobs are now validated up front and a bad one exits 2. `max=0` is rejected too, because it skipped the loop entirely and then reported `All 0 attempts failed` with exit 1 — indistinguishable from the command having run and failed.

The default schedule is unchanged. Verified with a `sleep` shim on `PATH`: it receives `5`, `10`, `20`.

## Why this survived review, and what the new tests assert instead

This is the part worth keeping. Every observable the existing tests watched sits upstream of the arithmetic:

- `test_bandit_scanner_behavior` asserts the config dict was passed through, with the installer stubbed out
- `test_uv_tool_mixin.test_with_custom_retry_config` mocks the runner wholesale and asserts `result is True`
- `TestUVToolRetryConfig` reads the dataclass's own fields back to itself
- every test in `test_with_retry.py` sets `WITH_RETRY_DELAY=0`, which collapses the backoff to a constant zero and makes the interval unobservable **by construction**
- the AWS retry tests pass `base_delay=0.01` to stay fast and then assert only the call count

None of those assertions was vacuous. They just watched something the defect does not move.

The 24 tests added here assert the things it does move: **the attempt count, the argv `sleep` received, and the exact interval schedule.** Jitter is switched off wherever an exact interval is asserted, so the schedule is arithmetic rather than a range. Nothing measures wall clock — a lower bound flakes on a fast machine and an upper bound on a slow one — so the shell test shims `sleep` onto `PATH` and reads back what it was handed.

## Mutation proof

Each fix was proved by mutating the specific thing it changes and confirming the right tests redden, and that the pre-existing ones do not:

| mutation | reddens | pre-existing tests |
|---|---|---|
| backoff made constant (`delay = base_delay`) | 3 interval tests | all 65 others pass |
| retry loop reverted (`attempts = 1`) | 6 tests | all 25 in the file pass |
| AWS exponential flattened + clamp removed | 4 tests | all 4 `TestRetryWithBackoff` pass |
| shell doubling made a no-op (`delay * 1`) | 1 sleep-argv test | all 7 others pass |

The constant-backoff mutation is the informative one: it reddens *only* the interval assertions while every attempt-count test stays green. Count assertions alone would have shipped it.

The flattened AWS mutation logs this, which is the defect shape verbatim — message printed, backoff destroyed:

```
WARNING Retrying due to ClientError (ThrottlingException). Attempt 1/2 after -30.00s delay
```

One negative result worth recording: removing the `max(0.0, ...)` clamp from `_install_retry_delay` initially reddened **nothing**, because `__post_init__` already normalizes the inputs. Two overlapping guards, neither distinguishable by test. Rather than delete one, there is now a test for the single case only the clamp can catch — `UVToolRetryConfig` is a plain mutable dataclass, so assigning `config.max_delay = -30.0` after construction bypasses normalization. That test fails with `[-30.0, -30.0, -30.0]` when the clamp is removed.

## A note on the sleep guard

The guard reads `if retry_config is not None and attempt + 1 < attempts:`. The first clause is redundant today — `attempts` is 1 whenever `retry_config` is falsy, so the second clause already prevents the call. It is written out anyway because a type checker cannot see that correlation, and more importantly because if a later change to how `attempts` is derived broke it, the failure would be an `AttributeError` raised *from inside the retry path*: code that runs only after something has already gone wrong, invisible to any test that never triggers a retry. Both halves are pinned — breaking the correlation reddens the attempt-count assertions, and removing the clause on top of that fires a sentinel in the delay helper.

## Follow-ups deliberately not in this PR

**The shell `retry` helper in `.github/actions/run-scan-test/action.yml` is sound and untouched.** It is owned by another open PR. Its `sleep $((n * 10))` is arithmetic expansion, not command substitution, and `n` is always a valid integer seeded at 1, so the delay cannot be empty, zero, or negative; the backoff grows 10s then 20s (linear, not exponential, but it grows). Read and left alone.

Two observations in that file, for whoever owns it. `ERROR_COUNT=$(echo "$REPORT_OUTPUT" | grep -c "ERROR" | head -1)` runs under the runner's default `-e -o pipefail`, and `grep -c` exits 1 when the count is zero — so a report containing no `ERROR` at all makes the assignment fail and kills the step. Separately, `FERRET_CONSTRAINT="$(...)"` is the command-substitution shape done correctly, guarded by an explicit `-z` check.

**Three more inert retry configs, same family, not fixed here.** `s3_reporter.py` and `bedrock_summary_reporter.py` expose `max_retries`/`base_delay`/`max_delay` in their user-facing config schema, and neither is read: `@retry_with_backoff()` at `s3_reporter.py:269` takes no arguments, and `bedrock_summary_reporter.py` imports `retry_with_backoff` and never calls it. `cloudwatch_logs_reporter.py:257` has the same bare-decorator gap for `put_log_events` while `:222` correctly forwards config.

Left out on purpose. These decorators are applied at class-definition time, so reading `self.config.options` requires restructuring each method to decorate an inner function at call time — a design change, not a defect fix. And the effective values currently equal the declared defaults (`3` / `1.0` / `60.0`), so nothing is broken today for anyone who has not overridden them. The defect is "a user override is silently ignored", not "no retry happens".

**Also noted, not changed:** `monitor_workspace_progress`'s `poll_interval` is an unvalidated public keyword parameter that busy-spins at `0` (no in-tree caller passes it). `_sleep_interval_for` is documented as a "Backoff schedule" but is a three-step ladder keyed to completed-scanner count, not attempt number, so it does not grow per attempt. `retry_with_backoff` has no upper bound on `max_retries`, and `base_delay * (2**retries)` would raise `OverflowError` somewhere past a thousand retries — unreachable in practice at a 60s cap, but the exponent is computed before the cap applies.

## Verification

- `pytest tests/unit`: **6520 passed, 24 skipped, 3 xfailed, 2 failed, 192 errors.** All 192 errors are one cause, `tests/unit/deploy - ModuleNotFoundError: No module named 'moto'`, repeated once per xdist worker — a missing optional test dependency in this environment. The 2 failures are `test_agent_plugin_ash_version.py::test_every_install_ref_names_the_packaged_version` and `test_cdk_nag_container_deps.py::test_fallback_matches_installed_metadata`, both version/dependency drift; neither file references any module this PR touches. Baseline is therefore 6496 passed with the same 2 failures and 192 errors.
- **No test was weakened, skipped, or deleted.** The test diff removes zero lines: +24 test functions, 0 removed, 0 existing assertions changed.
- `ruff check` on the five changed Python files: **83 findings, identical to the `origin/main` baseline** for the same files, with an identical rule breakdown. Zero new lint.
- `ruff format --check`: four of the five are clean, including `aws_utils.py`, which was *unformatted* at baseline — this change incidentally fixed that. The fifth, `test_uv_tool_runner.py`, was already unformatted on `main` under ruff 0.16 and the formatter's remaining diff touches none of the added lines. Note the tree is formatted for the ruff pinned in `.pre-commit-config.yaml` (v0.11.8), not the `>=0.16` in the dev dependency group.

### DIFF
--- a/automated_security_helper/assets/with-retry.sh
+++ b/automated_security_helper/assets/with-retry.sh
@@ -23,6 +23,41 @@ set -o pipefail
 # Overridable so the retry behaviour can be exercised without waiting out the
 # real backoff. Defaults are what every Dockerfile caller gets.
 max=${WITH_RETRY_MAX_ATTEMPTS:-3}; delay=${WITH_RETRY_DELAY:-5}; attempt=1
+
+# Both knobs feed integer arithmetic below -- `$((delay * 2))` and
+# `[ $attempt -le $max ]` -- and bash arithmetic cannot parse a non-integer.
+# A fractional WITH_RETRY_DELAY such as 0.1, which is the obvious way to ask for
+# a fast-but-nonzero backoff, made `delay=$((delay * 2))` a syntax error. That
+# error aborts the enclosing `while` WITHOUT aborting the script (there is no
+# `set -e` here, and adding one would change how the attempt loop reports), so
+# exactly ONE attempt ran and the script still printed "All 3 attempts failed"
+# and exited 1. The attempt count was wrong and the only message anybody reads
+# said otherwise, which is precisely why no assertion on that message could see
+# it. Reject the bad value instead of retrying once and misreporting it.
+#
+# `case` rather than a regex or `expr` so this stays dependency-free: it runs in
+# the container build stage before most of the toolchain exists.
+case $max in
+  '' | *[!0-9]*)
+    echo "with-retry: WITH_RETRY_MAX_ATTEMPTS must be a non-negative integer, got '$max'" >&2
+    exit 2
+    ;;
+esac
+case $delay in
+  '' | *[!0-9]*)
+    echo "with-retry: WITH_RETRY_DELAY must be a whole number of seconds, got '$delay'" >&2
+    exit 2
+    ;;
+esac
+# Guarded separately, and only once $max is known to be numeric so `-lt` is
+# safe. max=0 would skip the loop entirely and then report "All 0 attempts
+# failed" with exit 1 -- a failure indistinguishable from the command having
+# been run and failed, when it was never run at all.
+if [ "$max" -lt 1 ]; then
+  echo "with-retry: WITH_RETRY_MAX_ATTEMPTS must be at least 1, got '$max'" >&2
+  exit 2
+fi
+
 while [ $attempt -le $max ]; do
   bash -o pipefail -c "$*" && exit 0
   echo "Attempt $attempt/$max failed, retrying in ${delay}s..." >&2
--- a/automated_security_helper/plugin_modules/ash_aws_plugins/aws_utils.py
+++ b/automated_security_helper/plugin_modules/ash_aws_plugins/aws_utils.py
@@ -89,9 +89,30 @@ def retry_with_backoff(
                         # If we shouldn't retry, re-raise the exception
                         raise
 
-                    # Calculate delay with exponential backoff and jitter
-                    delay = min(
-                        base_delay * (2**retries) + random.uniform(0, 1), max_delay  # nosec B311 — jitter for retry backoff, not security
+                    # Calculate delay with exponential backoff and jitter.
+                    #
+                    # Clamped non-negative deliberately. ``base_delay`` and
+                    # ``max_delay`` reach here from reporter config --
+                    # CloudWatchLogsReporter and S3Reporter both declare them as
+                    # bare ``float`` fields with no lower bound and hand them
+                    # straight to this decorator -- so a negative ``max_delay``
+                    # made this expression negative, and ``time.sleep`` raises
+                    # ValueError on a negative argument. That ValueError is then
+                    # swallowed by the broad ``except Exception`` in
+                    # ``_create_log_stream_with_retry``, so the retry silently
+                    # became zero retries and the log blamed the log stream.
+                    #
+                    # The floor is 0.0 rather than some small positive number
+                    # because ``base_delay=0`` is a legitimate way to ask for
+                    # immediate retries; what must not happen is a *negative*
+                    # interval reaching sleep, or an exception escaping from the
+                    # arithmetic that is supposed to protect a throttled API.
+                    delay = max(
+                        0.0,
+                        min(
+                            base_delay * (2**retries) + random.uniform(0, 1),  # nosec B311 — jitter for retry backoff, not security
+                            max_delay,
+                        ),
                     )
 
                     # Log the retry
--- a/automated_security_helper/utils/uv_tool_runner.py
+++ b/automated_security_helper/utils/uv_tool_runner.py
@@ -1,5 +1,6 @@
 """UV tool runner utility for managing UV-based tool execution and installation."""
 
+import random
 import subprocess  # nosec B404 — uv_tool_runner is the subprocess orchestrator for tool execution
 import threading
 import time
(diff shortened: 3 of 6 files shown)


awsmadi:
Superseded by #560, which consolidates this with #551 and #552 under one theme: operations that claim to retry or cache actually do. All three retry defects land there unchanged -- the missing loop in `install_tool_with_version` (measured 1 attempt and 0 sleeps before, 4 and 3 after), `with-retry.sh` truncating its loop on a non-integer delay while claiming all three attempts failed, and the negative interval reaching `time.sleep`. #560 adds a test comparing the shell script's summary count against what actually ran, since the defect lived in the gap between those two observables. Branch kept.

PR #555, opened 2026-09-10
fix(cdk-nag): preserve pack attribution, separate unevaluated rules, and surface converted-target errors

awsmadi:
Three reporting defects on the cdk-nag path. None of them changes what the scanner finds; all
three change what the report says about what it found. Measured against the versions the `[cdk]`
extra installs today: **aws-cdk-lib 2.267.0, cdk-nag 3.0.2**.

Base: `origin/main` at `882b4b2a3bb9c0be3e046a5eab74eebcb16ec2bc`.

Every claim below was reproduced before it was fixed, and each fix has a test that fails on
`origin/main` and passes here. Failure text is quoted verbatim.

---

## Defect 1 — findings from a different tool were attributed to cdk-nag

**Verified.** With one important correction to the original report, in the last subsection.

### Mechanism

`validation-report.json` is CDK's *shared* policy-validation report. Every registered
`IPolicyValidationPlugin` writes into the same `pluginReports` array, and from aws-cdk-lib
2.262.0 the CDK registers one of its own unconditionally — `CloudFormationValidatePlugin`:

```js
// aws-cdk-lib 2.267.0, core/lib/validation/cloudformation-validate-plugin.js
class CloudFormationValidatePlugin {
  static PLUGIN_NAME = "CloudFormation Validate";
```

Its rules come from the bundled `@aws/cloudformation-validate` and are ids like `F3017`,
`W3010`, `W2001`, `E1151`. A live run through the wrapper against a template with a placeholder
KMS key identifier returns two packs in one report:

```
wrapper pack keys: ['AwsSolutions', 'CloudFormation Validate']
  pack 'AwsSolutions': 4 result(s)
  pack 'CloudFormation Validate': 3 result(s)
     ruleId='F3017' level=warning kind=informational
```

`level=warning` maps to MEDIUM on ASH's ladder, which is where the reported "16 MEDIUM `F3017`
findings" comes from — one violation fanning out across `violatingConstructs`.

The wrapper did key the packs apart, but the scanner discards the key:

```python
for pack, findings in nag_result_dict.results.items():
    ASH_LOGGER.debug(f"Found {len(findings)} findings for {pack} on template {cfn_file}")
    sarif_results.extend(findings)
```

`pack` is used for a debug line and nothing else. The pack name did reach
`result.properties.tags` as an unlabelled positional entry, but it never reached the **rule
descriptor** — the object a consumer reads to find out what a rule is and who owns it. Two
things were wrong there:

```python
helpUri=f"https://github.com/cdklabs/cdk-nag/blob/main/RULES.md#{str(finding_props.get('rule_level', 'rule')).lower()}s",
tags=finding_props.get("tags", []) + [ ... ],
```

* `helpUri` was cdk-nag's RULES.md for **every** rule. Following it for `F3017` opens a page
  that does not mention `F3017` and does not describe the mechanism that governs it.
* `finding_props` is `_NagFinding.as_dict()`, which has keys
  `rule_id / resource_id / compliance / exception_reason / rule_level / rule_info` — and no
  `tags` key. So `finding_props.get("tags", [])` returned its default on every finding ever
  scanned. It looked like it forwarded the finding's tags and forwarded none.

### Fix

The pack is recorded on each `_NagFinding`, travels through `as_dict()` into the SARIF property
bag, and is written onto the rule as a labelled `pack` property plus a `pack::` tag. `helpUri`
routes on it. Ownership is *derived* from cdk-nag's own id construction — `applyRule` builds
`f"{packName}-{ruleSuffix}"` — rather than from a list of pack names that would go stale
silently and start treating a new cdk-nag pack as foreign.

A finding with **no** pack recorded keeps its previous destination. Redirecting only findings
positively known to be foreign means missing data cannot move a help link; guessing "foreign"
from an absent pack would send genuine cdk-nag rules away from their own documentation, which is
the same misattribution in reverse.

The findings are **not** filtered out. They are real CloudFormation Validate findings and stay
in the report; only their attribution changes.

### Before / after

```
FAILED tests/unit/utils/test_cdk_nag_pack_attribution.py::TestOriginatingPackIsCarriedOnTheFinding::test_each_finding_records_the_plugin_that_produced_it - AttributeError: '_NagFinding' object has no attribute 'pack'
FAILED tests/unit/utils/test_cdk_nag_pack_attribution.py::TestOriginatingPackIsCarriedOnTheFinding::test_the_pack_survives_into_the_serialized_finding_record - KeyError: 'pack'
FAILED tests/integration/scanners/test_cdk_nag_real_pack.py::TestReportAttribution::test_a_cdk_plugin_that_is_not_a_nag_pack_is_named_as_its_own_pack - KeyError: 'pack'
FAILED tests/integration/scanners/test_cdk_nag_real_pack.py::TestReportAttribution::test_the_scanner_documents_a_foreign_rule_against_the_right_guide - KeyError: 'pack'
FAILED tests/unit/plugin_modules/ash_builtin/test_cdk_nag_scanner_behavior.py::test_rule_tags_include_the_findings_own_tags - AssertionError: the finding's own tags did not reach the rule; got ['tool_n...
```

All pass after. The integration tests run the real scanner against real cdk-nag, so the
end-to-end path is covered and not just the extracted helper.

### What a user sees differently

An `F3017` finding is labelled `pack: CloudFormation Validate` and tagged
`pack::CloudFormation Validate`, and its help link points at
`https://docs.aws.amazon.com/cdk/v2/guide/policy-validation-synthesis.html` — which documents
the plugin, the shared report, and `Validations.of(scope).acknowledge()`. Genuine cdk-nag rules
are unchanged and still point at RULES.md with the level-derived anchor. Rule descriptors now
carry the finding's own tags (resource id, resource type, pack) instead of only the two tool
tags.

### Correction: these findings were NOT unsuppressable

The original report said these 16 are "permanently unsuppressable" because a cdk-nag suppression
can never match them. **That part is not real, and I did not build anything on it.** Measured
directly: an in-template `Metadata.cdk_nag.rules_to_suppress` entry with `id: F3017` **does**
suppress the finding, because `_template_suppression_reason` matches the reported rule id
regardless of which plugin produced it.

```
pack='CloudFormation Validate' ruleId='F3017' level=warning suppressed=YES
    justification: (ASH cdk-nag in-template suppression) Placeholder KMS key identifier is intentional in this fixture.
```

So the real defect is the attribution, not the suppressibility. Worth knowing that the mechanism
which governs such a finding *upstream* is CDK's own — `Validations.of(scope).acknowledge({id:
'CloudFormation-Validate::F3017', reason: ...})`, stored as the `aws:cdk:acknowledged-rules`
construct metadata key — which ASH neither reads nor previously pointed the user toward. The new
`helpUri` is what points there. Reading that key is a separate feature and is not in this PR.

---

## Defect 2 — "rule could not be evaluated" was rendered as a violation

**Verified**, and worse than reported: the finding came out **CRITICAL**, not merely as a
violation.

### Mechanism

`CdkNagValidationFailure`, the 2.x concept, is **removed** in cdk-nag 3.x. In 3.0.2 the state
arrives with no structural marker at all — `applyRule` catches the exception and calls the *same*
`addViolation` a real violation goes through:

```js
// cdk-nag 3.0.2, lib/nag-pack.js
catch (error) {
    if (!this.isAcknowledged(params.node, ruleId)) {
        this.addViolation(ruleId, params, error.message);
    }
}
...
const description = errorMessage
    ? `Rule threw an error during validation. ${this.verbose ? errorMessage : 'This is generally caused by a parameter referencing an intrinsic function.'}`
```

The description prefix is the only thing separating the two. It is reached whenever a rule calls
`NagRules.resolveIfPrimitive` on a property that resolves to a non-primitive, which is what a
`Ref` to a CloudFormation Parameter resolves to:

```js
// cdk-nag 3.0.2, lib/nag-rules.js
if (resolvedValue === Object(resolvedValue)) {
    throw Error(`The parameter resolved to to a non-primitive value "...", therefore the rule could not be validated.`);
}
```

Reproduced with an EC2 SecurityGroup whose `GroupDescription` and an EBS Volume whose `Encrypted`
both come from parameters:

```
  pack 'AwsSolutions': 2 result(s)
     ruleId='AwsSolutions-EC27' level=error kind=fail
     ruleId='AwsSolutions-EC26' level=error kind=fail
  pluginName='AwsSolutions' violations=2
     ruleName='AwsSolutions-EC27' severity='error' desc='Rule threw an error during validation. The parameter resolved to to a '
```

`severity: "error"` → `_normalize_rule_level` → `"Error"` → `level=error, kind=fail`, which
`_SARIF_LEVEL_TO_SEVERITY` maps to **CRITICAL**. Two rules that never ran were two critical
security findings, and the coverage gap they represent was invisible.

### Fix — using SARIF's representation, not a new convention

SARIF 2.1.0 defines two candidate kinds and only one is a true statement. Quoting the OASIS
Standard incorporating Approved Errata 01, §3.27.9:

> `"notApplicable"` : The rule specified by ruleId was not evaluated, because it does not apply
> to the analysis target.

> `"open"` : The specified rule was evaluated, and the tool concluded that there was insufficient
> information to decide whether a problem exists.

`"open"` opens with "was evaluated", and its NOTE 1 scopes the value to proof-based tools that
completed an analysis and could not prove either direction. cdk-nag's rule raised partway
through, so "was not evaluated" is the accurate half. §3.27.9's own example for `notApplicable`
is a result whose message reads `"<target> was not evaluated for rule <id> because <reason>"` —
the shape emitted here.

`level` follows rather than being chosen. §3.27.10:

> `"none"` : The concept of "severity" does not apply to this result because the kind property
> (§3.27.9) has a value other than `"fail"`.

ASH maps `none` to INFO, so the count inflation stops as a consequence of getting the SARIF
right rather than as a separate adjustment that could drift apart from it.

The coverage gap is *also* reported at the run level, in the place SARIF defines for exactly
this. From the official schema: `invocation.toolExecutionNotifications` is "A list of runtime
conditions detected by the tool during the analysis", and `notification.associatedRule` is "A
reference used to locate the rule descriptor associated with this notification". Emitted at
`level=error` — overriding the field's `warning` default, because for a security scanner a rule
that silently did not run is more serious than a violation it did find — and deduplicated by
rule id, since one unresolvable property raises for every construct sharing its shape.

Detection anchors on the prefix rather than the full string (the remainder differs under
`verbose`), and matches at the start rather than anywhere in the text, so a rule whose own
explanation happens to mention an error is not demoted.

### Before / after

```
FAILED tests/unit/utils/test_cdk_nag_unevaluated_rule.py::...::test_end_to_end_the_unevaluated_rule_does_not_render_as_a_critical_failure - AssertionError: a rule cdk-nag could not evaluate rendered as an evaluated result; got (<Level.error: 'error'>, <Kind.fail: 'fail'>)
FAILED tests/integration/scanners/test_cdk_nag_real_pack.py::TestRuleThatCouldNotBeEvaluated::test_the_scanner_reports_an_unevaluated_rule_as_a_run_level_condition - AssertionError: the scan produced no toolExecutionNotifications, so a rule ...
```

The first is written deliberately without importing anything new, so its red state is a
behavioral difference rather than a missing symbol. Each test names the state that must be
present and, from the same report, a genuine violation that must stay `fail`/`error` — so a
change that demoted everything cannot pass.

### What a user sees differently

`AwsSolutions-EC26` on a parameter-driven property is no longer a CRITICAL finding. It appears as
`kind: notApplicable`, `level: none`, with the message `'<template>' was NOT evaluated for rule
AwsSolutions-EC26, so this result says nothing about whether the template complies with it.` The
run's invocation carries a `toolExecutionNotifications` entry at `error` level naming that rule
via `associatedRule`. A reader can now tell "this rule fired" from "this rule did not run", and
the finding count no longer includes rules that produced no verdict.

---

## Defect 3 — a converted target's ERROR rolled up to PASSED

**Verified.** It does fall through a gap between the two existing guards, as suspected.

### Mechanism

`ScanPhase` gives each scanner one task carrying both trees, and `ScanResultProcessor` writes one
serialized container per target under `additional_reports[scanner][target_type]`.
`get_scanner_status_info` derived its `error` flag from one report:

```python
elif (
    scanner_name in asharp_model.additional_reports
    and "source" in asharp_model.additional_reports[scanner_name]
    and asharp_model.additional_reports[scanner_name]["source"]["scanner_name"]
    and asharp_model.additional_reports[scanner_name]["source"]["status"]
):
    excluded, dependencies_missing, error = (
        ScannerStatisticsCalculator._status_info_from_report(
            asharp_model.additional_reports[scanner_name]["source"]
        )
    )
```

The `"converted"` entry is never read. That flag is what every rolled-up status keys on —
`get_unified_scanner_metrics` checks `stats["error"]` first, `get_scanner_status` returns
`"ERROR"` off the same tuple, and `ScannerMetrics.passed` is True for PASSED, so the row rendered
green *and* serialized as `passed: true` into `ash.flat.json`.

Both adjacent guards did their own job correctly, which is why this survived:

* `ScanResultsContainer.determine_status` returned ERROR for the converted container as designed.
  The loss is downstream of it.
* `scanner_evaluated_nothing` already reads across **every** target report and documents why —
  "only one of the two targets may have been empty". Correct quantifier, different fact. The
  asymmetry between those two functions is precisely where the defect lived.

### Fix

`error` is ORed across every target report via a new `any_target_errored`. `excluded` and
`dependencies_missing` deliberately stay on the single scanner-level report: both are facts about
configuration rather than about a target, and widening them would let one target with an
unrecognized status — which `_status_info_from_report` maps to `excluded=True` for backward
compatibility — claim the operator had switched the scanner off.

### Before / after

```
FAILED tests/unit/core/test_scanner_status_across_targets.py::...::test_status_info_reports_error_when_only_the_converted_target_errored - AssertionError: an ERROR on the converted target did not set the error flag...
FAILED tests/unit/core/test_scanner_status_across_targets.py::...::test_get_scanner_status_returns_error_for_a_converted_target_error - AssertionError: assert 'PASSED' == 'ERROR'
FAILED tests/unit/core/test_scanner_status_across_targets.py::...::test_unified_metrics_rolls_a_converted_target_error_up_to_error - AssertionError: assert 'PASSED' == 'ERROR'
```

`assert 'PASSED' == 'ERROR'` is the reported shape exactly. The mirror case (source ERROR,
converted PASSED) is asserted too, so the fix cannot be a source-for-converted swap, and three
parametrized negative controls (`PASSED`, `FAILED`, `SKIPPED` on the converted target) keep
`error = True` from being a valid implementation.

### What a user sees differently

`source PASSED, converted ERROR` now rolls up to `ERROR` with `passed: false`, in the summary
table, in `get_scanner_status`, and in `ash.flat.json`. A conversion step that failed outright is
visible instead of averaging away against a passing sibling.

---

## Checked against the live CI failure, which turns out to be something else

ASH's `scan-validation` and `install-validation` jobs are currently red with output that
looks like Defect 2 in production:

```
ERROR    cdk-nag failed to scan .../deploy/cdk/tsconfig.json: ScannerError: while scanning for the next token
ERROR    cdk-nag failed to scan .../mkdocs.yml: ConstructorError: could not determine a constructor for the tag
ERROR (2) Exiting due to 1 actionable findings found in ASH scan
```

The natural reading is that two parse failures became one actionable finding. **They did
not.** From the same job's own summary table (attempt 1, no re-run substitution):

```
┃ Scanner      ┃ S   ┃ C ┃ H ┃ M ┃ L ┃ I ┃ Time   ┃ Action ┃ Result  ┃ Thresh  ┃
│ cdk-nag      │ 437 │ 0 │ 0 │ 0 │ 0 │ 0 │ 1m 16s │ 0      │ PASSED  │ MED (g) │
│ detect-secr… │ 55  │ 1 │ 0 │ 0 │ 0 │ 0 │ 1m 0s  │ 1      │ FAILED  │ MED (g) │
```

The one actionable finding is a **detect-secrets CRITICAL**. cdk-nag is on `Action 0 /
PASSED`. Two unrelated facts printed near each other, which is why the arithmetic never
lined up.

Reproduced locally against the same two file shapes, on this branch:

```
targets_attempted = 3
targets_failed    = 2
SARIF results = 2
   by kind: {'fail': 2}
   results attributed to a target that failed to parse: []
exitCodeDescription (the error channel):
   .../mkdocs.yml: ConstructorError: could not determine a constructor for the tag ...
   .../tsconfig.json: ParserError: while parsing a flow mapping
```

A target-level parse failure produces **zero** findings and is recorded only in the error
channel. `actionable` is derived from SARIF severity counts, so it cannot receive one. The
error-to-finding conversion does not exist, so there is nothing to fix — and the fourth
commit pins that property, which nothing asserted before.

**Would this branch still print "1 actionable findings"? Yes, and correctly** — that count
is a genuine detect-secrets CRITICAL that should fail the build. Nothing here adds to or
removes from it. Two separate defects are visible in that job and neither is fixed here:

1. **Target selection hands cdk-nag files outside its domain.** `mkdocs.yml` and
   `tsconfig.json` are not CloudFormation. The `ConstructorError` / `ScannerError` are the
   YAML and JSON readers failing, upstream of anything attribution can help with.
2. **Partial coverage loss is invisible in the rolled-up status.** cdk-nag reported `PASSED`
   with `executionSuccessful: True` and `exitCode: 0` while 2 of 3 targets were never
   evaluated (in CI, also with a `cdk-nag produced no validation report at ...`).
   `determine_status` only reports ERROR once `targets_failed >= targets_attempted`, so
   *total* coverage loss is loud and *partial* coverage loss is silent. Defect 3 above is
   the adjacent case — an ERROR on one target being dropped — not this one.

Both deserve their own change. Neither is widened into this PR.

## Verification

**Full suite, as CI runs it** (`uv sync --extra cdk` then `uv run pytest`):

| | base `882b4b2a` | this branch |
|---|---|---|
| passed | 6571 | 6605 |
| failed | 1 | 1 |
| skipped | 119 | 124 |
| xfailed | 3 | 3 |

The single failure is identical on both sides and unrelated to this change:
`tests/unit/test_agent_plugin_ash_version.py::TestInstallRefsTrackPackagedVersion::test_every_install_ref_names_the_packaged_version`,
an install-reference/version-drift check. It is pre-existing on `main`.

The +5 skips are the five new integration tests, which the default invocation skips for every
integration test in the repository (`Need --run-integration option to run`). No pre-existing test
changed skip status. CI runs them explicitly, and so did I:

**Integration, as CI runs it**
(`ASH_REQUIRE_CDK_EXTRA=1 uv run pytest tests/integration/scanners/test_cdk_nag_real_pack.py --run-integration -p no:randomly --no-cov`):
**3 passed on base → 8 passed here.** The new tests were deliberately added to this existing file
rather than a new one, because that file path is what the CI step names and `.github/**` is owned
by another open PR.

Of the five, four fail on base for the behavioral reasons quoted above. The fifth
(`test_an_unparseable_target_is_an_error_not_a_finding`) passes on base too, and is labelled as
what it is: a regression test pinning a property that was already correct, added because the CI
output above made it look broken and nothing asserted it either way.

**Lint.** `ruff check --statistics` over the four modified files, run against the `origin/main`
blobs and the branch under the same ruff: **53 findings before, 53 after, zero delta on every
rule**. All 53 are pre-existing. Under ruff's documented default rule set the only findings are
two pre-existing unused imports, present identically on both sides. `ruff format --check` is
byte-identical per file before and after — the two files that were already non-clean on `main`
remain so with zero new hunks, and all five new files are clean.

**Python coverage.** 93.16% → **93.18%** (+0.02pp). Gate is 80%.

**No test was weakened, skipped, or deleted.** One test was rewritten:
`test_rule_tags_do_not_include_the_findings_own_tags` →
`test_rule_tags_include_the_findings_own_tags`. It was a characterization test whose own docstring
described the always-empty `finding_props.get("tags", [])` as a production defect ("So
`finding_props.get("tags", [])` is always empty in production and the resource/pack/type tags
never reach the rule"). The replacement is strictly stronger: it names every tag that must be
present instead of asserting `len(tags) == 2`. The `nag_result()` helper in the same file gained a
`pack` key so it keeps matching the shape the wrapper emits — that file already asserts the
fixture must do so.

### DIFF
--- a/automated_security_helper/core/scanner_statistics_calculator.py
+++ b/automated_security_helper/core/scanner_statistics_calculator.py
@@ -31,6 +31,7 @@ from typing import Dict, Any, Optional, Tuple
 
 from automated_security_helper.config.default_config import get_default_config
 from automated_security_helper.core.constants import ASH_DEFAULT_SEVERITY_LEVEL
+from automated_security_helper.core.enums import ScannerStatus
 from automated_security_helper.models.asharp_model import (
     AshAggregatedResults,
     ScannerSeverityCount,
@@ -592,6 +593,45 @@ class ScannerStatisticsCalculator:
         return excluded, dependencies_missing, error
 
     @staticmethod
+    def any_target_errored(
+        asharp_model: AshAggregatedResults, scanner_name: str
+    ) -> bool:
+        """True when ANY target this scanner reported under came back ERROR.
+
+        ``ScanPhase`` gives each scanner one task carrying both the source tree and the
+        converted tree, and ``ScanResultProcessor`` writes one serialized container per target
+        under ``additional_reports[scanner][target_type]``. ERROR is what
+        ``ScanResultsContainer.determine_status`` returns when a tracking scanner failed every
+        target it attempted -- so an ERROR here means no rule was evaluated against that tree.
+
+        ``get_scanner_status_info`` used to derive its ``error`` flag from the ``"source"``
+        report alone. A cdk-nag that passed on the source tree and errored on the converted one
+        therefore rolled up to PASSED: the ERROR container was written to disk, serialized into
+        the aggregated results, and never read. Nothing downstream could recover the fact,
+        because ``error`` is the flag every rolled-up status keys on.
+
+        "Any", not "all", and the asymmetry with the severity gate is the point. A finding count
+        of zero is ambiguous, so the guards in ``determine_status`` are careful about which
+        zero they are looking at. An ERROR is not ambiguous: one target that evaluated nothing
+        is a hole in the scan whether or not its sibling was fine, and a report that averages
+        it away against a passing sibling is asserting coverage it does not have.
+
+        Modelled on ``scanner_evaluated_nothing``, which already reads across every target for
+        the same structural reason. That function existed with that quantifier and this one did
+        not, and the gap between them is exactly where the defect lived.
+        """
+        reports = asharp_model.additional_reports.get(scanner_name)
+        if not isinstance(reports, dict):
+            return False
+
+        for target_report in reports.values():
+            if not isinstance(target_report, dict):
+                continue
+            if target_report.get("status") == ScannerStatus.ERROR.value:
+                return True
+        return False
+
+    @staticmethod
     def scanner_evaluated_nothing(
         asharp_model: AshAggregatedResults, scanner_name: str
     ) -> bool:
@@ -662,6 +702,15 @@ class ScannerStatisticsCalculator:
         genuinely different fact, and it is answered by ``scanner_evaluated_nothing`` rather than
         folded into ``excluded`` here -- see ``_status_info_from_report`` for what folding it in
         used to cost.
+
+        ``error`` is the one of the three read across every target rather than off a single
+        report, and that asymmetry is deliberate. A scanner is switched off, or its tool is
+        absent, for the whole run -- those are facts about configuration, and the scanner-level
+        report states them correctly. Whether a rule was evaluated is a fact about one tree, and
+        a scanner gets two. See ``any_target_errored`` for what reading only the source report
+        cost. Widening ``excluded`` the same way would be actively wrong: one target carrying an
+        unrecognized status makes ``_status_info_from_report`` return ``excluded=True`` for
+        backward compatibility, which would then claim the operator had switched the scanner off.
         """
         excluded = False
         dependencies_missing = False
@@ -703,6 +752,14 @@ class ScannerStatisticsCalculator:
         else:
             # If the scanner is not found in the dictionary, check for errors
             error = True
+
+        # ORed in after the branches above rather than replacing them. Those branches also
+        # produce ``excluded`` and ``dependencies_missing``, which stay scanner-level, and the
+        # first branch reads a report keyed ``"None"`` -- a scanner that never reached a target
+        # at all -- which ``any_target_errored`` covers as well since it walks every key.
+        error = error or ScannerStatisticsCalculator.any_target_errored(
+            asharp_model, scanner_name
+        )
 
         return excluded, dependencies_missing, error
 
--- a/automated_security_helper/plugin_modules/ash_builtin/scanners/cdk_nag_scanner.py
+++ b/automated_security_helper/plugin_modules/ash_builtin/scanners/cdk_nag_scanner.py
@@ -17,9 +17,15 @@ from automated_security_helper.core.exceptions import ScannerError
 from automated_security_helper.schemas.sarif_schema_model import (
     ArtifactLocation,
     Invocation,
+    Level,
+    Message,
+    Message1,
     MultiformatMessageString,
+    Notification,
     PropertyBag,
     ReportingDescriptor,
+    ReportingDescriptorReference,
+    ReportingDescriptorReference3,
     Result,
     Run,
     SarifReport,
@@ -98,8 +104,8 @@ def _rule_help_uri(rule_id: str, pack: str, rule_level: str) -> str:
 
     The CDK's policy-validation guide is the correct destination for those: it documents the
     plugin, the shared validation report, and the ``Validations.of(scope).acknowledge()``
-    mechanism that governs a CloudFormation Validate finding. It is also the right destination
-    for a third-party ``IPolicyValidationPlugin``, since that guide is what defines the protocol
+    mechanism that governs a CloudFormation Validate finding. It is also the right fallback for
+    a third-party ``IPolicyValidationPlugin``, since that guide is what defines the protocol
     such a plugin implements.
 
     A finding with NO pack recorded keeps the previous destination, and that is deliberately
@@ -113,6 +119,71 @@ def _rule_help_uri(rule_id: str, pack: str, rule_level: str) -> str:
     if not pack or _rule_is_from_pack(rule_id, pack):
         return f"{_CDK_NAG_RULES_URL}#{str(rule_level).lower()}s"
     return _CDK_POLICY_VALIDATION_URL
+
+
+def _unevaluated_rule_notifications(
+    results: list[Result],
+) -> list[Notification]:
+    """One SARIF notification per rule that could not be evaluated.
+
+    WHY THE RESULT ALONE IS NOT ENOUGH
+    ----------------------------------
+    A ``notApplicable`` result says "this one rule reached no verdict on this one resource",
+    and that is true but easy to miss: it sits in the same results array as the findings, at a
+    severity of ``none``, and most report surfaces sort or filter by severity. The fact a reader
+    needs is coarser -- "part of this scan did not run" -- and it belongs where a reader looks
+    for facts about the run.
+
+    SARIF has exactly that place, and the schema says so in its own words. ``invocation``'s
+    ``toolExecutionNotifications`` is "A list of runtime conditions detected by the tool during
+    the analysis", and ``notification.associatedRule`` is "A reference used to locate the rule
+    descriptor associated with this notification". A rule raising mid-evaluation is a runtime
+    condition, and the rule it happened to is the thing to associate it with. This is the
+    representation SARIF already defines for the case, so it is used rather than a bespoke
+    property.
+
+    ``level=error`` rather than ``warning``. For a security scanner, a rule that silently did
+    not run is the more serious of the two facts it can report -- a violation at least tells you
+    what to fix. ``warning`` is the field's default, so this is a deliberate override.
+
+    Deduplicated by rule id. One rule that cannot resolve a property will raise for every
+    construct that shares the shape, and the run-level statement is about the rule, not about
+    each occurrence -- the per-resource detail is already carried by the results themselves.
+    """
+    from automated_security_helper.utils.cdk_nag_wrapper import NOT_EVALUATED
+
+    notifications: list[Notification] = []
+    seen: set[str] = set()
+    for result in results:
+        finding_props = (result.properties.model_extra or {}).get("cdk_nag_finding", {})
+        if finding_props.get("compliance") != NOT_EVALUATED:
+            continue
+        rule_id = result.ruleId or ""
+        if rule_id in seen:
+            continue
+        seen.add(rule_id)
+        pack = str(finding_props.get("pack", "") or "unknown")
+        notifications.append(
+            Notification(
+                level=Level.error,
+                message=Message(
+                    root=Message1(
+                        text=(
+                            f"Rule {rule_id} from pack '{pack}' could not be evaluated, so "
+                            "this scan reports nothing about compliance with it. "
+                            f"{finding_props.get('rule_info', '')}".strip()
+                        )
+                    )
+                ),
+                associatedRule=ReportingDescriptorReference(
+                    # The id-bearing variant. ReportingDescriptorReference is a union whose
+                    # other two members require an index or a guid, neither of which exists
+                    # here -- the rule is identified by the id cdk-nag reported it under.
+                    root=ReportingDescriptorReference3(id=rule_id)
+                ),
+            )
+        )
+    return notifications
 
 
 def _reporting_descriptor_for(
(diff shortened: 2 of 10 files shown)


awsmadi:
Consolidated into #559, which carries these four commits unchanged alongside the partial-coverage reporting work from #557 and one behavior change built on both. The cdk-nag reporting fixes and the coverage-visibility work turned out to be the same theme — a scanner's report telling the truth about what it actually scanned — and the new gate change depends on both halves, so reviewing them as one branch is more useful than reviewing them apart. Branch kept, not deleted.

PR #557, opened 2026-09-10
feat(reporting): surface how much of its input each scanner evaluated

awsmadi:
A scanner that evaluated one file in three reports a clean scan, and nothing in the output says
two thirds went unchecked. This makes the coverage legible. It deliberately does **not** change
any status or exit code — that part is a policy decision and is handed back below.

Base: `origin/main` at `882b4b2a3bb9c0be3e046a5eab74eebcb16ec2bc`.

## The behavior, measured across the range

Real `ash scan` runs, cdk-nag isolated with `--exclude-scanners`, four source trees of exactly
three targets each so only `targets_failed` varies. `--no-fail-on-findings` so the exit code
reflects coverage rather than the findings that happen to come back from the targets that worked.

| targets failing | `determine_status` | `targets_failed` | SARIF `executionSuccessful` | exit (default) | exit (`--fail-on-incomplete-scanners`) | summary status | actionable | `incomplete_scanners()` |
|---|---|---|---|---|---|---|---|---|
| 0/3 | PASSED | 0 | True | 0 | 0 | PASSED | 6 | none |
| 1/3 | PASSED | 1 | True | 0 | 0 | PASSED | 4 | none |
| 2/3 | PASSED | 2 | True | 0 | 0 | PASSED | 2 | none |
| 3/3 | ERROR | 3 | False | 0 | **1** | ERROR | 0 | `cdk-nag=ERROR` |

The cliff is exactly at "all failed". 1/3 and 2/3 are indistinguishable from 0/3 in every status
channel.

Three things worth naming that the table shows:

1. **`scanner_evaluated_nothing` does not partly handle this.** It keys on
   `targets_attempted <= 0`; three targets were attempted in every row.
2. **The existing opt-in completeness gate is blind to it too.**
   `_INCOMPLETE_SCANNER_STATUSES` is `{ERROR, MISSING}` and `incomplete_scanners()` selects on
   status, so `--fail-on-incomplete-scanners` sees only the 3/3 row. The mechanism built for
   "tell me when scanners did not run" does not cover "a scanner ran on a third of its input".
3. **With the default gate off, even 3/3 exits 0.** The status, the summary table and the SARIF
   all correctly say ERROR / `executionSuccessful: False`, and the process still exits 0 because
   `fail_on_incomplete_scanners` defaults off and there are no findings to trip the other gate.
   That is documented default behavior rather than a new defect, but it means total coverage loss
   is loud in the report and silent in the exit code.

One measurement was confounded and is reported as such: running the same matrix with `--scanners
cdk-nag` instead of `--exclude-scanners` made **all four** rows exit 1, because an allowlist does
not mark the unselected scanners as excluded and their absent tools report MISSING on their own.
That is a documented limitation of `incomplete_scanners()`, not data about coverage, so the table
above uses `--exclude-scanners` as that docstring directs.

## Blast radius, measured on this repository

cdk-nag over ASH's own tree: **10 targets attempted, 4 failed — 40%** — reported `PASSED`, with
163 findings from the 6 that succeeded.

The 4 are not one kind of thing, and that is the crux:

- `deploy/cdk/tsconfig.json` — JSON with comments and backticks; never a template
- `mkdocs.yml` — `!!python/name:pymdownx.emoji.twemoji`; never a template
- `tests/.../test-yaml.template.json` — a real template, `cdk-nag produced no validation report`
- `tests/.../cfn-and-python-test-yaml.template.json` — same

So two of the four are routine noise and two are **CloudFormation templates that genuinely went
unscanned**. The signal is hidden among the noise. A file that parses and simply is not
CloudFormation is *not* counted as a failure — the wrapper returns None and the scanner decrements
`targets_attempted` back — so this population is specifically "could not be parsed" plus "no
report produced", which is exactly the set a reader needs to see.

## Were the counts visible anywhere?

Present in `ash_aggregated_results.json` and `scanners/<name>/<target>/ASH.ScanResults.json` as
raw counters; the failed filenames reach `ash.log` and both SARIF files' `exitCodeDescription`.

Absent from **`ScannerMetrics`** — the single model the summary table and every reporter read. Its
fields were suppressed / critical / high / medium / low / info / scanner_name / actionable /
duration / status / threshold / threshold_source / excluded / dependencies_missing / passed. No
target counters, so the summary table, `ash.flat.json` and the markdown, HTML and text reporters
could not show them even in principle. That is the half of the defect that is purely additive.

Adding the model field was **not** sufficient, and assuming it would be was wrong.
`ReportContentEmitter.get_scanner_results()` builds its dict by enumerating keys explicitly, and
flatjson, markdown, HTML and text all go through it. Measured after adding the field and before
touching the emitter: `targets_attempted` still absent from a real `ash.flat.json`. There is a
test for that specifically, because the mistake was mine.

## What this PR changes

- `ScannerMetrics.targets_attempted` / `.targets_failed`, summed across **every** target report
  rather than read off `"source"` — a scanner gets two targets and either can lose coverage.
  `targets_attempted` keeps the container's tri-state meaning: `None` means the scanner does not
  track targets (bandit, checkov, semgrep, grype, syft, detect-secrets, opengrep, cfn-nag,
  npm-audit) and must not collapse to `0`, which is the different claim "tracked and attempted
  none".
- `coverage_shortfalls()` — `(scanner, attempted, failed)` for scanners that lost input, so the
  counts are assertable as data rather than as prose.
- Both counters in `get_scanner_results()`, so they reach `ash.flat.json` and every reporter built
  on it. `null` rather than `0` for a non-tracking scanner.
- A console line after the summary table, and a section after the markdown summary's scanner
  table, naming each scanner and its counts. Emitted only when something was lost.

Verified end to end on a 2-of-3 tree:

```
Incomplete coverage: cdk-nag evaluated 1 of 3 target(s); 2 could not be evaluated ...

### Incomplete coverage
- **cdk-nag**: evaluated 1 of 3 target(s); 2 could not be evaluated
```

Beside the table rather than a twelfth column: `len(table.columns) == 11` is asserted, the
narrow-terminal layout already carries eleven, consumers parse it positionally, and a shortfall is
exceptional rather than per-row. The markdown section is **not** suppressed in compact mode —
compact drops noise, and this is signal in the rendering most likely to be pasted into a PR.

## The design fork, handed back

`ERROR on any failed target` is **clearly wrong** and I am not proposing it. At the measured 40%
rate it turns a routine condition into a permanent red for any repository containing a `mkdocs.yml`
with a pymdownx tag or a `tsconfig.json` with comments, and it conflates the two routine parse
failures with the two genuinely-unscanned templates. That one does not need a decision.

The remaining options all turn on how you want pipelines to behave, so none is implemented:

| option | effect | blast radius |
|---|---|---|
| **E. Teach `incomplete_scanners()` about partial coverage** | The existing `--fail-on-incomplete-scanners` opt-in starts covering 1/3 and 2/3, not just 3/3 | Default unchanged. Anyone already using that flag flips from green to exit 1 — including ASH's own repo at 40%. Smallest new surface: no enum change, no new config, reuses the switch that already means "tell me when scanners did not run" |
| **B. A distinct status (e.g. `PARTIAL`)** | Visible in the status column itself, and gateable | New `ScannerStatus` member touches every status consumer — reporters, junitxml, ocsf, the table, `_INCOMPLETE_SCANNER_STATUSES`, and the `passed` property, whose value for it is the same policy question again. Downstream parsers see an unknown status |
| **D. A configurable threshold** | Fail above N% or N targets unevaluated | New config surface, and its default *is* the policy question. Most flexible, most to document, and silently does nothing until someone sets it |

**My recommendation if you want a failure mode: E.** It reuses an opt-in that already expresses
exactly this intent, changes no default, and adds no schema. The cost is real and worth stating
plainly: operators who set `--fail-on-incomplete-scanners` today would start failing on
repositories they currently pass on, which is why it is a decision and not a fix.

Whichever way that goes, the visibility in this PR stands on its own and does not need to be
revisited.

## Verification

Both directions. The new tests on `origin/main`: **9 failed, 3 passed** — behavioral reds
`AttributeError: 'ScannerMetrics' object has no attribute 'targets_attempted'` and
`KeyError: 'targets_attempted'`, the rest missing symbols. After: **18 passed**.

The 3 that pass on both sides are the deliberate controls asserting the status and `passed` are
**unchanged** (`PASSED`/`True` for partial, `ERROR`/`False` for total). They are labelled as such:
they exist so a later change to the verdict has to edit them.

Per-assertion discipline, since a coverage fix is easy to fake:

- Counts asserted as exact totals through `coverage_shortfalls`, never "greater than zero" —
  reading only the source report gives `(3, 2)` and only the converted `(1, 0)`, and a loose
  assertion accepts both while dropping half the input.
- Renderers asserted for emit-versus-silence only, never for wording.
- A negative control per channel: nothing printed, and no markdown section, when no coverage was
  lost. Without those, a fix that warned on every scan would pass.
- `None`-not-`0` asserted for non-tracking scanners in both the metrics and the emitted dict.
- The markdown scanner table's column count asserted, so the shortfall cannot have been
  implemented as an extra column after all.

**Full suite**, as CI runs it (`uv sync --extra cdk` then `uv run pytest`):

| | base `882b4b2a` | this branch |
|---|---|---|
| passed | 6571 | 6589 |
| failed | 1 | 1 |
| skipped | 119 | 119 |
| xfailed | 3 | 3 |

Skips unchanged — no integration tests added. The single failure is identical on both sides and
unrelated: `test_every_install_ref_names_the_packaged_version`, an install-reference/version-drift
check, addressed in a separate PR.

**Lint.** `ruff check --statistics` over the four modified files against the `origin/main` blobs
under the same ruff: **30 before, 30 after, zero delta on every rule.** The new test file has zero
findings. `ruff format`: **zero new hunk lines** in any modified file (three already drift on
`main` and drift identically here); the new file is clean.

**Python coverage.** 93.16% → **93.18%**.

No test weakened, skipped or deleted. No file touched that the cdk-nag reporting PR also touches —
that PR owns `scanner_statistics_calculator.py`, `cdk_nag_scanner.py` and `cdk_nag_wrapper.py`, and
this one owns `unified_metrics.py`, `metrics_table.py`, `report_content_emitter.py` and
`markdown_reporter.py`. The intersection is empty, so the two can land in either order.

One thing deliberately left out for that reason: a SARIF notification per unevaluated target would
belong in `cdk_nag_scanner.py`'s invocation, which the other PR already edits to add
`toolExecutionNotifications` for a related purpose. Adding a second writer to the same object from
a second branch buys a conflict for no benefit, so it belongs on top of that PR rather than beside
it.

### DIFF
--- a/automated_security_helper/core/metrics_table.py
+++ b/automated_security_helper/core/metrics_table.py
@@ -8,6 +8,7 @@ from rich.panel import Panel
 
 from automated_security_helper.models.asharp_model import AshAggregatedResults
 from automated_security_helper.core.unified_metrics import (
+    coverage_shortfalls,
     get_unified_scanner_metrics,
     format_duration,
 )
@@ -156,6 +157,41 @@ def generate_metrics_table_from_unified_data(
     return table
 
 
+def print_coverage_shortfalls(
+    asharp_model: AshAggregatedResults, console: Console
+) -> None:
+    """Name every scanner that could not evaluate part of its input, with the counts.
+
+    Printed BESIDE the table rather than added to it. A twelfth column would change the table's
+    geometry for every scan including the ones with nothing to report, and the responsive
+    narrow-terminal layout already carries eleven; the shortfall is also exceptional rather than
+    per-row data. Emitting nothing when there is nothing to say is what keeps this from becoming
+    noise -- see the negative control in
+    ``tests/unit/core/test_partial_target_coverage_visibility.py``.
+
+    Why it exists at all: ``determine_status`` only returns ERROR once every attempted target
+    failed, so a scanner that lost some of its input reports whatever the severity gate gives it.
+    Measured on this repository, cdk-nag attempts 10 targets, fails 4, and reports PASSED -- and
+    two of those four were real CloudFormation templates rather than files that were never
+    templates. Nothing in the rendered output said so.
+
+    The status and the exit code are deliberately not touched. At the measured rate, failing on
+    any unevaluated target would turn a routine condition into a permanent red; whether it should
+    fail is a policy question about pipelines and not one this rendering decides.
+    """
+    shortfalls = coverage_shortfalls(asharp_model)
+    if not shortfalls:
+        return
+    for scanner_name, attempted, failed in shortfalls:
+        evaluated = attempted - failed
+        console.print(
+            f"[bold yellow]Incomplete coverage:[/bold yellow] {scanner_name} evaluated "
+            f"{evaluated} of {attempted} target(s); {failed} could not be evaluated and "
+            f"contributed no findings. See the scanner's exitCodeDescription in its SARIF "
+            f"report for the reason per target."
+        )
+
+
 def display_metrics_table(
     asharp_model: AshAggregatedResults,
     source_dir: str | Path | None = None,
@@ -248,6 +284,10 @@ def display_metrics_table(
             console.print()
             console.print(table)
             console.print()
+            # After the table, so the counts sit next to the status they qualify. Inside the same
+            # guarded block because it writes to the same console: a closed stdout has to be
+            # handled identically for both, and a shortfall notice is not worth failing a scan.
+            print_coverage_shortfalls(asharp_model, console)
         except (ValueError, OSError, BrokenPipeError) as io_error:
             # stdout/stderr may be closed (e.g., in MCP server context)
             # Log the error but don't fail the scan
--- a/automated_security_helper/core/unified_metrics.py
+++ b/automated_security_helper/core/unified_metrics.py
@@ -74,6 +74,26 @@ class ScannerMetrics(ScannerSeverityCount):
     excluded: bool = False  # Whether the scanner was explicitly excluded
     dependencies_missing: bool = False  # Whether the scanner has missing dependencies
 
+    # How much of its input the scanner actually evaluated, summed over every target it reported
+    # under. Present here because this class is the only model the summary table and every
+    # reporter read: the counters existed on ``ScanResultsContainer`` and in the serialized
+    # per-target reports, and had no route to any human-facing output at all.
+    #
+    # That mattered because ``determine_status`` only returns ERROR once
+    # ``targets_failed >= targets_attempted``. A scanner that failed on some of its targets keeps
+    # whatever status the severity gate gives it, so partial coverage loss renders exactly like a
+    # clean scan. Measured against cdk-nag on this repository: 10 targets attempted, 4 failed,
+    # reported PASSED -- and two of those four were CloudFormation templates that genuinely went
+    # unscanned rather than files that were never templates.
+    #
+    # ``targets_attempted`` keeps the container's tri-state meaning rather than defaulting to 0.
+    # None means the scanner does not track per-target outcomes and is making no claim; bandit,
+    # checkov, semgrep, grype, syft, detect-secrets, opengrep, cfn-nag and npm-audit are all in
+    # that state. Reading them as "attempted zero" would report every one of them as having
+    # evaluated nothing.
+    targets_attempted: int | None = None
+    targets_failed: int = 0
+
     @computed_field
     @property
     def passed(self) -> bool:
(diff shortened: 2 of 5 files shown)


awsmadi:
Consolidated into #559, which carries this commit unchanged alongside the cdk-nag reporting fixes from #555. This PR deliberately left the exit code alone and said so; #559 now makes that decision explicitly, teaching --fail-on-incomplete-scanners about the partial-coverage case these counters made visible, so the counts and the gate that reads them land together rather than one without the other. Branch kept, not deleted.

PR #559, opened 2026-09-10
fix(reporting): make a scanner's report tell the truth about what it scanned

awsmadi:
Consolidates #555 and #557 and adds one behavior change on top. The three parts share a theme: **a scanner's report should tell the truth about what it actually scanned.** Each is a separate commit; the file sets of the two original PRs are disjoint, so this is a clean combine rather than a merge resolution.

## 1. cdk-nag attributed findings to the wrong thing, and reported unevaluated rules as critical

Three defects in how cdk-nag findings were reported, all of them in the SARIF the scanner produces.

**Pack attribution was lost at the rule descriptor.** The reporting descriptor read `finding_props.get("tags", [])`, and `_NagFinding.as_dict()` has no `tags` key. It therefore returned `[]` on *every finding ever scanned*, and `helpUri` pointed at cdk-nag's own `RULES.md` for every rule regardless of which pack raised it. A finding from AwsSolutions and a finding from HIPAA were indistinguishable in the report, and the help pointer led somewhere that does not document the rule.

**A rule that could not be evaluated was reported as a critical finding.** `CdkNagValidationFailure` no longer exists in v3. A rule that throws during validation now arrives through `applyRule`'s catch calling `addViolation`, distinguishable only by the message prefix `"Rule threw an error during validation."` Those were being emitted as `level=error, kind=fail` — CRITICAL on ASH's severity ladder — when what they mean is "this rule did not evaluate." They now report `kind=notApplicable` with `level=none`, which the SARIF spec requires once `kind != fail`. A genuine violation still reports `level=error, kind=fail`, and there is a test holding that line so the fix cannot quietly reclassify real findings.

**An error on any scanned target now reaches the scanner status,** instead of being lost when only one of a scanner's two targets failed.

Adds `NOT_EVALUATED`, `_violations_from_validation_report`, `_level_and_kind`, `_unevaluated_rule_notifications`, `_rule_is_from_pack`, and `_reporting_descriptor_for`.

## 2. A scan could not say how much of its input it evaluated

`ScannerMetrics` is the only model the summary table and every reporter read, and it had no field for either target counter. The counts existed on the container, in `ash_aggregated_results.json`, and in the scanner's own SARIF — with no route to any human-facing output even in principle.

Adds `targets_attempted`/`targets_failed` preserving a tri-state `None`, plus `coverage_shortfalls()`, surfaced through `metrics_table.py`, `report_content_emitter.py` and `markdown_reporter.py`.

Two things worth knowing before changing this code:

- **Adding the model field was not sufficient.** `ReportContentEmitter.get_scanner_results()` builds its dict by enumerating keys explicitly, so a new model field reaches `ash.flat.json` only if it is listed there too. Measured: `targets_attempted` was absent from a real `ash.flat.json` with the model field already populated. I swept the other consumers on this path for the same bug class — `_populate_summary_stats_from_unified_metrics` and `_populate_scanner_results_from_unified_metrics` are both hand-built enumerations, but their target models (`SummaryStats`, `ScannerTargetStatusInfo`) carry no target counters at all, so there was nothing for them to drop.
- **The shortfall is reported beside the summary table, not as a twelfth column.** `test_the_scanner_table_keeps_its_column_count` pins `len(table.columns) == 11` deliberately, because consumers parse that table positionally.

## 3. `--fail-on-incomplete-scanners` was blind to partial coverage loss

The new behavior change, and the reason the first two belong together.

The gate selected on scanner *status*:

```python
_INCOMPLETE_SCANNER_STATUSES = frozenset({ERROR, MISSING})
```

and `determine_status` returns `ERROR` only once `targets_failed >= targets_attempted`. Those two facts compose into a hole: a scanner that failed on *some* of its targets keeps whatever status the severity gate gives it, so the gate never saw it. The flag reported total coverage loss and stayed silent on partial loss — which is both the more common case and the one operators turn the flag on to catch.

### Behavior change

**This turns an existing exit 0 into an exit 1 for some users with no change to their own code.** It affects only runs that pass `--fail-on-incomplete-scanners` or set `fail_on_incomplete_scanners: true`. The default path is unchanged, and there is a test asserting that the same partial-coverage input still exits 0 without the flag. Recorded in `CHANGELOG.md` rather than left for people to discover.

Measured blast radius on this repository's own tree: **exactly one scanner is newly caught.** cdk-nag attempts 10 targets and cannot evaluate 4 of them (40%), so ASH's own scan flips to exit 1 under the flag. That is the intended behavior, not a regression.

Two of the four are real CloudFormation templates that went unscanned, and two are a target-selection bug:

| target | cause | genuinely a template? |
|---|---|---|
| `tests/.../test-yaml.template.json` | cdk-nag produced no validation report | **yes — unscanned** |
| `tests/.../cfn-and-python-test-yaml.template.json` | cdk-nag produced no validation report | **yes — unscanned** |
| `deploy/cdk/tsconfig.json` | `ScannerError: while scanning for the next token` | no, never a template |
| `mkdocs.yml` | `ConstructorError` on a `pymdownx.emoji.twemoji` tag | no, never a template |

So the 40% is not entirely spurious: half of it is real coverage loss. Fixing target selection is deliberately **out of scope here**.

### Deliberately not done

- **No new `ScannerStatus` member and no new config key.** This is the existing opt-in flag learning about a case it was always meant to cover. A partially-covered scanner still reports its own status, so no reporter, no summary table, and no consumer of the status enum reads differently.
- **No status-shaped expression of partial coverage,** and that is a caller constraint rather than taste. `cli/merge.py::_completed` keys on status against `_INCOMPLETE_SCANNER_STATUSES` to answer a narrower question — whether a shard's scanner ran *at all* — and `_verify_shard_contributions` refuses the merge outright where a shard completed none of the scanners it owned. A scanner that lost one target of ten did run, so a status for partial coverage would propagate into shard refusal and fail merges far from the cause. That is the concrete cost of adding a `ScannerStatus` member here, and the reason none was added.

  Worth stating precisely, because my first version of this reasoning overstated it and the code comment now records the correction. `_completed` inspects a `ScannerTargetStatusInfo`, which declares no target counters, so it cannot read coverage today by any route — the protection is structural, not merely conventional. Confirmed by mutation: reimplementing `_completed` to consult `_partial_coverage` changes nothing, because the value it would consult is never there. The boundary is trustworthy because the status does not change, which `test_partial_coverage_loss_still_reports_passed` pins directly, rather than because of the frozenset. `TestShardContributionIsRefused::test_a_shard_that_lost_only_some_targets_is_still_merged` pins it from the merge side, and that test's falsifiable half is the only coverage anywhere that the counters survive a merge — they live in `additional_reports`, which the merge recombines from n shards — and still reach the gate afterwards.
- **No tolerance threshold.** A percentage floor would need a number nobody can justify, and "1 of 200 templates went unscanned" is still a template nobody looked at.

### The tri-state is the subtle part

`targets_attempted` is `None` for the nine scanners that do not track per-target outcomes (bandit, checkov, semgrep, grype, syft, detect-secrets, opengrep, cfn-nag, npm-audit). `None` means "makes no claim"; `0` means "tracked, attempted none". Reading absence as "attempted 0, therefore lost everything" would fail every scan on every repository the moment the flag was turned on — the exact inverse of the defect being fixed. All three values are pinned end to end, and confirmed on a real `ash.flat.json`, where the nine non-tracking scanners serialize `null` rather than `0`.

Both counters are type-checked rather than trusted, matching what `unified_metrics.target_counts` already does to the serialized counters.

The operator-facing message said the scanners "did not run", which is false for the new case — the scanner ran and reported a status. Both the scan and merge messages now say the scanners did not evaluate everything they were given, and the guidance names the target-count case so nobody is sent to install a tool that is already installed.

## Verification

Against an `origin/main` baseline captured first, because main is currently red on one unrelated test.

- **Baseline (`origin/main`):** 1 failed, 6571 passed, 119 skipped, 3 xfailed. The failure is `test_agent_plugin_ash_version.py::TestInstallRefsTrackPackagedVersion::test_every_install_ref_names_the_packaged_version`, which another PR fixes.
- **This branch:** 1 failed, 6648 passed, 124 skipped, 3 xfailed — the same single inherited failure, **no new failures**. Confirmed on CI across Linux, Windows and macOS separately, each reporting exactly one failure and the same test, so a platform-specific new failure is not hiding behind a matching total.
- **The gate change was proven able to fail.** Run against the unmodified code first, the new tests failed 6 of 20, all six being the partial-coverage cases; the 14 controls (default path, tri-state negatives, total-loss message preservation) passed before and after. Whole file green after the change.
- **Coverage, per language, unblended:** Python 93.20% (gate is 80%). TypeScript, both trees: `deploy/cdk-constructs` 97.35% statements over 116 tests in 3 suites; `deploy/cdk` 94.36% statements over 355 tests in 13 suites. This change touches no TypeScript.
- **`ash config lint`** clean. **`ruff check`** byte-identical to the `origin/main` baseline on a per-rule statistics diff. The repo formatter was run on the changed files, and the three that remain unformatted are unformatted on main too.
- **Measured on a real artifact rather than the model in isolation:** a live cdk-nag scan of this repository, whose `ash.flat.json` carries `targets_attempted: 10` and `targets_failed: 4` for cdk-nag, and the exit code flipping 0 → 1 on those real results with and without the flag.

One coverage gap in the new code was found and closed rather than left: `_partial_coverage` guards its two counters independently, and every existing case was rejected on the *attempt* count and returned before the failure count was examined, leaving that second guard present, plausible, and never executed. It is now parametrized over the four shapes a malformed failure count takes.

### DIFF
--- a/.ash/.ash.yaml
+++ b/.ash/.ash.yaml
@@ -2085,6 +2085,29 @@ global_settings:
         of the literals they describe. Deliberately not citing a line number: this
         file argues against that a few entries above, and the number had already
         gone stale by the next commit.
+    - rule_id: SECRET-HEX-HIGH-ENTROPY-STRING
+      path: deploy/cdk-constructs/test/ash-scan-step.test.ts
+      reason: >-
+        The same 40-character git-ref placeholder as the entry above, reached by a
+        second file. Here it is the `commit` const in the jest test "PIP pins the
+        given git ref", which passes it to ASHScanStep as `version` and asserts
+        the synthesized buildspec carries it into the pip install URL. It stands
+        in for a commit rather than naming one -- the digits 0-9 followed by a-f,
+        repeated to length -- so it identifies nothing, grants nothing, and there
+        is nothing to revoke.
+        HexHighEntropyString fires on entropy alone here, with no keyword or
+        pattern component to corroborate it: the string spends all sixteen hex
+        digits across 40 characters, putting its Shannon entropy at 3.97 against
+        the plugin's limit of 3.0. Measured, not inferred -- any 40-character hex
+        string clears that limit, which is why a git SHA and a key are
+        indistinguishable to this detector.
+        A second entry rather than a widened first one because the two files sit
+        in different packages, deploy/cdk and deploy/cdk-constructs, and neither
+        path pattern reaches the other. One path per file and no
+        `deploy/cdk-constructs/**` glob, so a hex string added elsewhere in that
+        package still fails loudly instead of inheriting this argument.
+        Value DESCRIBED rather than quoted, and no line number cited, for the two
+        reasons the entry above records from having got both of those wrong.
     # === semgrep/opengrep Python compatibility (ASH requires 3.10+) ===
     - rule_id: "python.lang.compatibility.*"
       path: "automated_security_helper/**/*.py"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,34 @@
 - [1.0.1-e-10Jan2023](#101-e-10jan2023)
 
 
+## Unreleased
+
+### Behavior changes
+
+- **`--fail-on-incomplete-scanners` now also fails a scan that lost only part of
+  its input.** The flag selected on scanner status, and a scanner that failed on
+  some of its targets keeps the status the severity gate gives it — normally
+  `PASSED` — because `determine_status` returns `ERROR` only once
+  `targets_failed >= targets_attempted`. The flag therefore reported total
+  coverage loss and stayed silent on partial loss, which is the more common case
+  and the one operators turn it on to catch.
+
+  **This can turn an existing exit 0 into an exit 1 with no change to your own
+  code.** It affects only runs that pass `--fail-on-incomplete-scanners` (or set
+  `fail_on_incomplete_scanners: true`); the default path is unchanged. On this
+  repository's own tree the flag now exits 1, because cdk-nag attempts 10 targets
+  and cannot evaluate 4 of them.
+
+  A scanner that reports no target counts at all is unaffected — absent counters
+  mean the scanner does not track targets, not that it lost them, so the nine
+  scanners in that state cannot trip the gate. Scanner statuses themselves are
+  unchanged, so no report or summary table reads differently.
+
+  To restore the previous behavior, drop the flag (or set
+  `fail_on_incomplete_scanners: false`) to accept a partial scan. To keep the
+  flag and clear the failure, fix or exclude the targets the scanner could not
+  read; the failure message names each scanner with the counts.
+
 ## v3.7.0 (2026-08-27)
 
 ### Feat
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,9 +169,13 @@
     `ash.ghas.sarif` has no `invocations` key at all. A pipeline that uploads
     `ash.sarif` itself rather than `ash.ghas.sarif` will surface them.
   - A cdk-nag rule that could not be evaluated is now `kind: notApplicable` at
-    `level: none`, which ASH maps to INFO. It was previously reported as a
-    finding at the rule's declared severity, so some scans will show fewer
-    critical cdk-nag findings and gain the same number of informational ones.
+    `level: none`, which ASH maps to INFO. It was previously reported as a finding
+    at the rule's declared severity. Measured on this repository, the aggregated
+    SARIF holds the same 453 results before and after, of which 15 move to
+    `notApplicable`/`none` — 10 that were `error`/`fail` and 5 that were
+    `warning`/`informational`. The actionable counts do not move here because those
+    15 were already suppressed; on a tree where they are not, the scanner's
+    high-severity count drops by however many of its rules raised.
 
 ## v3.7.0 (2026-08-27)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,11 +56,26 @@
   coverage loss and stayed silent on partial loss, which is the more common case
   and the one operators turn it on to catch.
 
-  **This can turn an existing exit 0 into an exit 1 with no change to your own
-  code.** It affects only runs that pass `--fail-on-incomplete-scanners` (or set
-  `fail_on_incomplete_scanners: true`); the default path is unchanged. On this
-  repository's own tree the flag now exits 1, because cdk-nag attempts 10 targets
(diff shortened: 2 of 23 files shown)


nikmeiser (dismissed):
**cdk-nag rule-evaluation errors move from exit-2 to exit-0 by default, with no gate covering them**
(`cdk_nag_wrapper.py:257-259`; scanner counters `cdk_nag_scanner.py:520-547`). A rule that threw during validation was formerly `error/fail `→ CRITICAL → exit 2; it is now `notApplicable/none` → INFO → not actionable.

A rule-throw on an otherwise-parseable template does not increment
`targets_failed` (that rises only on a missing report or scan exception), so `--fail-on-incomplete-scanners` doesn't catch it either. The only remaining signals are a WARNING log and a run-level `toolExecutionNotification`, neither of which reaches GHAS or the exit code. Net: a scan whose only defect is systematic rule-evaluation failure
flips from exit-2 to clean exit-0 under default config, with no gate.

awsmadi:
Fixed in 75acf564 — deliberately not as implied, so the deviations are visible:

1. It exits 1, not 2. Exit 2 asserts that clearing the listed findings clears the scan; an unevaluated rule is not a finding you can clear.
2. It is not mapped to a severity, because SARIF forbids it: §3.27.10 requires `level: none` whenever `kind` is anything but `fail`, and `fail` asserts the rule *was* evaluated and the target failed. A severity-bearing result would claim a verdict never reached, untrue in the opposite direction.
3. It does not use `targets_failed`. `determine_status` returns ERROR once `targets_failed >= targets_attempted`, so a rule throwing on every template would report "No rules were evaluated" about a run that evaluated all but one.
4. Instead, `_compute_exit_code` reads `invocation.toolExecutionNotifications` at `level: error` — SARIF's own channel for a runtime condition detected during analysis, which the scanner already writes one of per unevaluated rule. Default on, ahead of the `fail_on_findings` early return.
5. It is suppressable, and that was load-bearing: the first version was not, and would have broken this repo's own CI with no escape hatch, since `.ash/.ash.yaml` already carries reviewed suppressions for rules that throw. A rule is reported only when at least one of its *not-evaluated* results is unsuppressed, filtered on `kind`.

GHAS still will not see these, and the scanner's own `executionSuccessful`/`exitCode` are untouched: both follow from the spec-correct `notApplicable`/`none`, and are separate judgments.

Approval was dismissed by the push (`dismiss_stale_reviews_on_push` on `main`), not by me. Re-requesting.

nikmeiser (approved):

PR #564, opened 2026-09-11
fix(exit-code): an incomplete scan is not a clean scan

awsmadi:
## This deliberately turns green CI cells red

Read that first, because it is the point of the change rather than a side effect. Several matrix cells that have been passing were passing without measuring what they claimed to measure. They now fail, and each of those failures is correct.

## The defect

ASH's exit code was derived from finding counts alone. Zero findings is what "nothing was wrong" and "nothing was checked" both look like, so a scan that could not run the scanners it was configured to run exited 0 — the same code as a scan where everything ran and found nothing.

Measured on pull request #546 at head `7d3759d3`, on check runs that were **green**:

| Check run | Scanners MISSING |
| --- | --- |
| `scan (python-local, windows-latest)` | 4 — cfn-nag, grype, semgrep, syft |
| `scan (python-local, ubuntu-24.04-arm)` | 3 — cfn-nag, grype, syft |
| `scan (python-local, ubuntu-latest)` | 3 — cfn-nag, grype, syft |
| `scan (python-local, macos-latest)` | 3 — cfn-nag, grype, syft |

Every missing scanner recorded under a millisecond, so none of them started. In the cells where semgrep did run it reported 82 findings — the green Windows cell with 4 of 10 scanners contributing nothing was not clean, it was unmeasured. The step that ran the scan recorded `outcome=success`, because ASH's own exit code did not reflect a scanner failure at all.

MISSING did not trip the plugin-error guard while ERROR did, which is why the four-MISSING cell stayed green.

Two more measurements from the same investigation:

- One of the five per-method CI guards read `scanners[*].result`. That field does not exist. In PowerShell's non-strict mode it resolved to `$null`, the comparison was vacuous, and the guard printed success unconditionally. Five per-method guards are now one script.
- `SummaryStats` had no `error` counter. On a run with two ERROR scanners, `passed + failed + missing + skipped` summed to 8 of 10 and every one of those totals read clean. The five counters now partition the set. The exit code does not read them — it reads per-scanner status, which is what lets it name the offender — but anything else deriving a verdict from the tallies was reading a total that silently excluded the worst outcomes.

## The exit-code contract

`1` means the scan did not happen. `2` means there are actionable findings. They stay distinguishable, and when both hold `1` wins: reporting `2` tells a reviewer that clearing the listed findings clears the scan, and that is false when a scanner contributed nothing.

Measured on a host without cfn-nag, grype and syft, against a one-file fixture:

| Invocation | Before | After |
| --- | --- | --- |
| `ash scan` (3 of 10 MISSING) | exit 0 | exit 1, naming all three |
| `ash scan --no-fail-on-incomplete-scanners` | exit 0 | exit 0 |
| `ash scan --scanners bandit` | exit 0 | exit 0, other nine SKIPPED |
| findings present, every selected scanner ran | exit 2 | exit 2 |

`SKIPPED` deliberately does not trip the gate. A skipped scanner is one you did not select, and that is also how sharding divides work — each shard excludes the scanners its siblings own. Gating on SKIPPED would fail every shard of a healthy sharded scan.

## Which cells this reddens, and why each is correct

The four `scan (python-local, *)` cells above now fail wherever their tools are genuinely absent. That is the intended outcome: those runs were already not measuring what they claimed, and the green check run is how nobody found out.

119 other cells were fixed by **declaring their scanner sets** rather than by opting out of the gate: the `install-validation` matrix is 5 operating systems x 4 Python versions x 8 install methods less 44 exclusions, so 116 cells, plus 3 `upgrade-path` cells. Every self-test scan in those jobs now passes `--scanners bandit`. That is a statement about what the jobs test, not a way to quieten them — they install ASH and nothing else, no `ash dependencies install`, no `[cdk]` extra, no Ruby gem, no Go binaries, so bandit is the only default scanner whose tool is present. bandit specifically because it is a core dependency in `pyproject.toml`. The nine scanners left out are recorded SKIPPED, so the report says what was and was not measured.

`--no-fail-on-incomplete-scanners` was **rejected** for those cells. It would have kept the same green while asserting nothing about whether any scanner ran, which is the property that made this defect invisible in the first place. An absent tool has two honest answers: install it, or exclude the scanner so the report says it was not part of the run. `false` is a third answer that returns to a 0 carrying no information. The flag appears nowhere in `.github/` except in the comment recording why it was not used.

## Workspace mode

The workspace layer mirrored only half of `_compute_exit_code`. It reproduced the threshold pass — same severity precedence, same suppression skip, same `--min-severity` gate — and not the completeness pass. So `ash --source-dir P` exited 1 on a host missing P's scanners while the same project inside a workspace exited 0.

A project that completed with a scanner at ERROR or MISSING is now reported with `scan_incomplete: true` and the scanner named in `incomplete_scanners`, and the run exits 1.

Workspace mode keeps its own contract rather than borrowing single-project precedence, and the difference is deliberate on both sides. An incomplete project lands on `INTERNAL_ERROR` (1), which in workspace mode sits **below** `ACTIONABLE_FINDINGS` (2) — the reverse of single-project mode. A workspace can hold an incomplete project and a project with real findings at the same time, so leading with the unknown would let one project's absent tool suppress another project's certainty. A single-project scan has no second project to suppress. Exit code 4 is untouched: it means no project was attempted at all.

Two smaller things in the same area:

- The completeness check delegates to the existing `incomplete_scanners` rather than mirroring it. That function reads through the unified metrics, where an excluded scanner is SKIPPED even when its tool is also absent; read straight off `scanner_results` the same entry is MISSING. A copy would have failed a workspace for a scanner the operator excluded. Two tests hold that distinction and were checked to discriminate — with the naive implementation substituted, they are the only two of thirteen that fail.
- The workspace rollup wrote `dependencies_satisfied: true` and `excluded: false` as constants next to a status of MISSING. `ScannerStatisticsCalculator` derives "this scanner is missing" from `dependencies_satisfied`, not from `status`, so anything re-deriving a verdict from a workspace results file read a MISSING scanner as present. Both are now derived from the rolled-up status.

## cfn-nag is now installed, and it has findings

`.ash/.ash.yaml` sets `cfn-nag: enabled: true`, but the scan workflow installed grype, syft and opengrep and not cfn-nag. The scanner resolved MISSING on every run, so the CloudFormation in this repository has never been checked by it. With the change above, ASH correctly reports its own scan incomplete — so the workflow now installs cfn-nag.

The version is not repeated in the workflow. `automated_security_helper/assets/Gemfile` and its committed `Gemfile.lock` are the pin the container image already builds from, and the step installs from those two files. No `ruby/setup-ruby` step is needed: `ubuntu-latest` installs Ruby from apt's `ruby-full`, which depends on `ruby-dev`, and the image also carries gcc, make and libyaml-dev — everything the two native gems in this lock (bigdecimal and psych 3.x) need in order to build.

Running ASH with cfn-nag available against this tree reports **56 findings, none suppressed, all actionable at the repository's MEDIUM threshold**. All of them are in the five committed CDK-synthesized templates under `deploy/cdk/templates/`:

| Count | Rule | Message |
| --- | --- | --- |
| 13 | CFN_NAG_W12 | IAM policy should not allow * resource |
| 12 | CFN_NAG_W84 | CloudWatchLogs LogGroup should specify a KMS Key Id to encrypt the log data |
| 8 | CFN_NAG_W76 | SPCM for IAM policy document is higher than 25 |
| 4 | CFN_NAG_W58 | Lambda functions require permission to write CloudWatch Logs |
| 4 | CFN_NAG_W77 | Secrets Manager Secret should explicitly specify KmsKeyId |
| 4 | CFN_NAG_W89 | Lambda functions should be deployed inside a VPC |
| 4 | CFN_NAG_W92 | Lambda functions should define ReservedConcurrentExecutions |
| 2 | CFN_NAG_W33 | EC2 Subnet should not have MapPublicIpOnLaunch set to true |
| 2 | CFN_NAG_W35 | S3 Bucket should have access logging configured |
| 1 | CFN_NAG_W5 | Security Groups found with cidr open to world on egress |
| 1 | CFN_NAG_W40 | Security Groups egress with an IpProtocol of -1 found |
| 1 | CFN_NAG_W56 | Elastic Load Balancer V2 Listener Protocol should use HTTPS for ALBs |

By file: `AshDistributedPipeline` 16, `AshFargate` 16, `AshCodeCommitGate` 11, `AshAgentCore` 9, `AshImagePipeline` 4.

These are left as findings. Nothing is suppressed and the severity threshold is unchanged — whether to fix them, suppress them with justification, or accept them is a separate decision from making the scanner run at all. So `SAST, SCA, and IaC Scan` will report exit 2 on findings until that decision is made.

## Verification

- `uv sync --extra cdk` then `uv run pytest`: **4807 passed, 115 skipped, 0 failed**. Baseline before the workspace change was 4794 passed, 115 skipped; the difference is the 13 new tests added here.
- New tests were shown red before green. `TestScannerCompleteness` failed with `assert 0 == WorkspaceExitCode.INTERNAL_ERROR` on a project reporting no findings and one MISSING scanner.
- `ruff format --check`: 200 files would be reformatted on both sides of the change, identical sets, nothing new.
- `ruff check`: no new findings in the rule set the repository enforces. Three new findings appear under a newer ruff's extended defaults (two UP006, one UP045), from using `List[...]` and `Optional[...]` to match the surrounding files.


awsmadi:
Closing this in favor of three focused pull requests. The work is preserved in full, not dropped.

## Why

This branch had grown to 111 commits — 22 of them merges, 2 reverts — across 24 distinct commit scopes and 74 files, spanning at least five unrelated concerns: CDK template size and suppression reasons, cdk-nag scanner status reporting, scan exit-code semantics, CI gate wiring, and Terraform encryption-at-rest. It was an integration branch rather than a reviewable change, and it had gone `CONFLICTING` against `main` after #559 landed.

An adversarial review of the branch found two real defects, both at seams between otherwise careful pieces of work. That is the cost of a change nobody can review in one sitting, and it is the reason for splitting rather than continuing to rebase.

## The successors

| PR | Files | Scope |
|---|---|---|
| #574 | 19 | CDK template size and suppression reasons, plus the pre-commit rule that exists for the committed templates |
| #576 | 7 | cdk-nag scanner status truthfulness |
| #575 | 50 | Scan exit-code semantics, scanner selection, CI completeness gates |

Together they contain the entire delta between `main` at `cc20a9938d` and this branch's merged state: 19 + 7 + 50 = 76 files, with no file unclaimed and none claimed twice. Each file was taken byte-identical from the merged tree and verified two ways — `git diff --exit-code` and per-file blob hashes — before and after committing.

## Notes for reviewers

**#576 depends on #575 and should merge after it.** #575 inverts a hand-maintained denylist of scanner statuses into a fail-closed allowlist, renaming the exported symbol. Two of #576's tests import that symbol, so #576 shows 2 collection errors until #575 lands. This is stated in #576's own description; it is an ordering constraint, not a defect.

**A cherry-pick split was not possible.** With 22 merge commits the history is not linear, and cherry-picking across merges can silently drop or duplicate changes while still producing commits that compile. The successors were therefore rebuilt from file state, with byte-equality assertions standing in for history. The commit messages here remain the record of why each individual change was made, which is part of why this stays closed rather than deleted.

**One premise in this branch's older commit messages went stale.** They describe the two inline-launchable templates as exceeding CloudFormation's 51,200-byte inline cap. On current `main` both are already under it, by 266 and 486 bytes. #574 widens thin headroom to 2,391 and 4,928 bytes; it does not repair a breach. #574's description says so.

PR #572, opened 2026-09-15
fix(cfn-nag): count failed templates so a broken scan cannot report PASSED

arpitjain099:
`cfn_nag_scanner.py` builds a local `failed_files` list, appends to it on both failure paths, and never reads it. `grep -n failed_files` on the file returns exactly those three lines. It also never sets `targets_attempted` or `targets_failed`, so `determine_status` has nothing per-target to look at and falls through to the finding count.

The consequence is that cfn-nag failing on every template is reported as a clean scan. With a stub `cfn_nag_scan` on PATH that runs, exits 1 and writes nothing:

```
$ ash --source-dir src --output-dir out --scanners cfn-nag
│ cfn-nag  │ 0 │ 0 │ 0 │ 0 │ 0 │ 0 │ 10ms │ 0 │ PASSED │ MED (g) │
$ echo $?
0
```

and in `ash_aggregated_results.json`:

```json
{"scanner_name": "cfn-nag", "target_type": "source", "exit_code": 1,
 "finding_count": 0, "status": "PASSED", ...}
```

The scanner's own exit code of 1 is recorded right next to the PASSED. Nothing downstream can tell that from a repository whose templates have no findings.

cdk-nag had exactly this and it is already fixed. `cdk_nag_scanner.py:456-461` carries the note:

> The counters set at the top of this method replace a local `failed_files` list that was appended to on both failure paths and never read, so a run that failed on every template still produced an empty-but-successful report.

and `ScanResultsContainer.record_target_failure` states the contract directly:

> Callers must use this rather than a local list. A local accumulator is invisible to status computation, so appending to one and never reading it produces a scanner that fails on every target and still reports success.

So this is the same fix applied to the scanner that was left behind, in the same shape:

- `targets_attempted` and `targets_failed` initialized at the top of `scan()`, above all five early returns, for the reason the cdk-nag comment gives.
- One attempt counted per file, after the `get_model_from_template` check, so a YAML file that is not a template stays an expected skip rather than an attempt.
- One failure counted for empty stdout and for stdout that does not parse as SARIF, with the file and reason appended to `self.errors`.
- The empty-stdout log moved from `debug` to `error`. At debug it was invisible on a default run, which is how this stayed quiet.
- An error line when every attempted template failed, wording matched to cdk-nag's.

After the change the same stub run gives:

```
│ cfn-nag  │ 0 │ 0 │ 0 │ 0 │ 0 │ 0 │ 63ms │ 0 │ ERROR │ MED (g) │
```

I did not touch how ERROR maps to the ASH process exit code; that is the same before and after.

### Tests

Five cases added to `test_cfn_nag_scanner_behavior.py`, each asserting the counters and then running the two lines from `ScannerExecutor` that copy them into a container, so a change at that boundary fails here too:

- empty stdout on the only template: attempted 1, failed 1, `scan_succeeded` false, status ERROR, and the "NOT a clean scan" line is logged
- stdout that is not SARIF: same
- usable output: attempted 1, failed 0, the finding survives, `scan_succeeded` true
- one of two templates failing: attempted 2, failed 1, `scan_succeeded` true
- a non-CloudFormation YAML file: attempted 0, failed 0, the subprocess is never called

All five fail on main and pass with the change. `pytest tests/unit` gives 156 failures on my checkout both before and after, the same 156 by name (`moto` and other absent test dependencies), so the change introduces none.

One note on an existing test: `test_empty_stdout_yields_no_findings_and_no_exception` is documented as "cfn_nag exiting 0 with no output is recorded, not treated as clean" but only asserts that a debug line was logged. It still passes unchanged.


nikmeiser (dismissed):

nikmeiser:
Thanks for this — it's approved and all required checks are green, but it can't enter the merge queue. GitHub reports **"Rebase conflict between head and base."**

That's not a content conflict — a plain merge/squash of this branch is clean. The cause is branch *shape*: the branch carries two `Merge branch 'main'` commits, and the merge queue rebases a PR's commits to form its queue entry. Rebase can't replay merge commits, so it surfaces as a "rebase conflict."

Could you linearize the branch? From your fork clone:

```
git checkout fix/cfn-nag-failed-targets
git fetch upstream                 # the remote pointing at awslabs/automated-security-helper
git rebase upstream/main           # flattens the two merge commits into your single commit
git push --force-with-lease
```

I verified locally that this rebase applies cleanly (exit 0) and collapses the branch to just your one commit — `fix(cfn-nag): count failed templates so a broken scan cannot report PASSED` — with no conflicts. Once the branch is linear, it'll be queue-able. Thanks!

awsmadi (dismissed):
lgtm! Thank you for your contribution!

awsmadi (approved):
lgtm

awsmadi:
Both Windows failures are fixed and CI is green here: 252 checks passing, none failing.

The first was `[WinError 193] %1 is not a valid Win32 application` from `cfn_nag_scan`. RubyGems writes two files for a gem executable, an extensionless binstub and a `.bat` wrapper, and only the wrapper is executable by `CreateProcess`. `find_executable` looked for the bare name and `.exe` only, so on Windows it returned the binstub. It now tries `.exe`, `.bat`, `.cmd`, then the bare name, which also fixes `npm audit` resolving to `npm.cmd`.

The second was a `charmap` codec error reading the results file. `automated_security_helper/cli/report.py` opened it without an encoding, so it decoded as cp1252 on Windows. That is now explicit UTF-8, along with the same pattern in `cli/mcp/progress_monitor.py` and `cli/mcp_server.py`.

Two notes on the state of this PR. The earlier approvals were dismissed by my push, not withdrawn by the reviewers — the branch protection dismisses stale reviews on push, and the fixes above needed the extra files. And the remaining block is the required CodeQL check, which has no result on this PR: no CodeQL analysis exists for any of its commits, so the required tool cannot report a threshold either way. That needs a maintainer rather than another push.

PR #575, opened 2026-09-15
fix(scan): the exit code has to be able to say "nothing was checked"

awsmadi:
Third of three PRs replacing #564, which is 111 commits across 24 scopes and cannot be cherry-picked. #574 is the first (cdk deploy templates); the cdk-nag scanner work is the second. This one carries the exit-code and CI-gate work. #564 stays open until all three verify.

Rebuilt from file state rather than by cherry-pick: branched off `cc20a9938d`, then checked out exactly 50 paths from a frozen source commit and asserted every one byte-identical, before and after committing — `git diff --exit-code` over the set returned 0 bytes, and all 50 blob hashes match. `git diff --name-only cc20a9938d HEAD` equals the 50 claimed paths exactly, so nothing outside them moved.

## Two defects, one shape

A scan that did not measure something reported itself clean, and every gate agreed.

### A run where no scanner reached a verdict exited 0

`_compute_exit_code` judged completeness one scanner at a time. SKIPPED is on the completeness allowlist and has to stay there — it is how sharding and `--exclude-scanners` record work a run was never meant to do — so a results file in which *every* entry is SKIPPED cleared the per-scanner pass having measured nothing.

Reachable from one word, with no sharding and no unusual config. On Windows, with this repository's own `.ash/.ash.yaml` (which sets `semgrep: enabled: true`), `ash scan --scanners semgrep` resolves semgrep against the registered names, so `ScannerSelectionError` does not fire; semgrep declares the platform unsupported and lands on SKIPPED; the other nine are not selected and land there too; findings are 0; exit 0.

Measured end to end through the shipped CLI on a one-file fixture, using `--exclude-scanners` over all ten to reproduce the all-SKIPPED shape on Linux:

| invocation | result |
|---|---|
| all excluded, new check disabled | `ash` exit **0**, ten SKIPPED, findings 0 |
| the same results file, CI gate | `assert_scanners_completed.py` exit **1** |
| all excluded, new check enabled | `ash` exit **1**, message names every status |
| `--scanners bandit` (control) | `ash` exit **0**, bandit PASSED beside nine SKIPPED |

`.github/scripts/assert_scanners_completed.py` already asserted this and caught it in CI. Its docstring claimed that reading the JSON status means the guard and ASH's own exit code "answer from the same field, so they cannot disagree" — and on that input they did, because only the script made the set-level assertion. `_RAN_SCANNER_STATUSES` is spelled identically to the script's `RAN_STATUSES` and a test holds the two together, since the script cannot import ASH: the bash and PowerShell scan methods leave no ASH on the runner's PATH. ERROR is excluded from the ran-set even though an ERROR scanner did run, because it reached no verdict, and because any ERROR entry already fails the script's per-scanner pass — counting it as having run would recreate the disagreement.

Two carve-outs, both pinned by tests. One shard of a split scan is excused: `core.sharding` allows a shard count above the scanner count so a pipeline can parameterise it without knowing how many scanners exist, and a shard owning one platform-declined scanner is the same shape. The union is not excused — `cli.merge._merged_exit_code` builds its `ScanOptions` with no shard fields, so the merged model is held to the assertion. An empty scanner set is a separate condition and deliberately not covered: `ash scan --phases convert` records no scanner at all, and the CI script refuses an empty `scanner_results` explicitly.

The check sits **inside** the `fail_on_incomplete_scanners` gate, and that was forced rather than chosen. `test_cli_false_overrides_config_true` already pins that with the gate off, a run whose only scanner is MISSING exits 0 — a run that measured nothing just as thoroughly. Placing this one check outside would answer one question two different ways depending on which status the non-running scanners happened to land on. The gate defaults on, so the case above fails by default. The script's docstring now names that as the one remaining difference instead of claiming there is none.

### A scanner that crashed during filtering was recorded as one nobody selected

`ScanPhase`'s filtering loop ends in `except Exception: continue`, so a scanner whose filtering raised leaves it in none of `enabled_scanner_names`, `excluded_scanner_names` or `dependency_error_scanners`, and with no `scanner_results` entry. The `getattr` guard added for `unsupported_platform_reason` covers a plugin that does not *have* the hook, not one whose hook *raises* — which a third-party scanner's can, for reasons of its own. The escape is older and wider than that hook.

What that produced is not the absence it looks like, and the difference is the point. Measured both ways:

| harness | recorded |
|---|---|
| validation manager mocked | no `scanner_results` entry at all |
| real `ScannerValidationManager` | `status=SKIPPED excluded=True dependencies_satisfied=True` |

The second is what a scan gets. `_validate_result_completeness` calls `ensure_complete_results`, which backfills an entry for every registered scanner missing from the results; `validate_scanner_enablement` has by then labelled the escaped scanner `enablement_status="disabled"` ("reason unknown"), `determine_scanner_status_from_execution_data` maps `"disabled"` to `"excluded"`, and `_create_missing_scanner_result_entry` writes the deliberate-deselection triple byte for byte. That is worse than no entry: SKIPPED is on the completeness allowlist, so `incomplete_scanners()` returned `[]` and the CI gate tolerated it. The absence is only what a *test* sees, because the existing suite's `_classify()` replaces the validation manager with a `MagicMock` — a harness that mocks the component deciding the behavior.

Closed by reconciling `scanner_instances` against `scanner_results` plus the queued set after the loop, which covers every route through that `except` rather than only the hook. Recorded MISSING, the only status meaning "was selected and did not run"; ERROR would claim it ran and failed. `excluded=False` is the load-bearing field — it is what stops `ensure_complete_results` taking its "properly excluded" branch and reasserting SKIPPED. `ScanResultsContainer.for_missing_deps` is deliberately not used despite being the usual source of this status: it hard-codes `dependencies_satisfied=False`, and nothing here checked a dependency, so False would tell an operator to install a tool that may be present. The resulting `(MISSING, deps=True, excluded=False)` triple is produced by no other path, so a results file carrying one says which of the three it was.

## Mutation control per new assertion

Each reverted and restored. One of these found a defect in a test rather than in the code.

| mutation | result |
|---|---|
| set-level check disabled | 3 red: all-SKIPPED, the Windows shape, the non-shard control. Every negative control stayed green. |
| SKIPPED added to the ran-set | 6 red, including the CI-script agreement test — the widening mutation is caught |
| shard carve-out removed | 3 red: exactly the three shard parametrizations |
| empty-set guard removed | 2 red: the boundary test and the unit test |
| reconciliation disabled | 5 red, and the two messages differ by harness, which is what established the mechanism above |
| accounted set drops the queued names | the healthy-scanner control **passed** — the executor double overwrites the entry moments later, so two unrelated suites were the only things red. The control now asserts the dispatch-time snapshot and fails on that mutation. |

## Verification

- `pytest`: **6851 passed, 127 skipped, 3 xfailed, 0 failed, exit 0**, coverage 93.23%.
- Run under `uv sync --extra cdk --frozen`. A plain `uv sync` installs no optional extras, and `tests/unit/utils/test_cdk_nag_unevaluated_rule.py:339` is a bare `import cdk_nag` with no `importorskip` and no `skipif`, so it fails rather than skips without the extra. That is pre-existing on `cc20a9938d` and untouched here.
- `ruff`: zero findings added and zero removed against `cc20a9938d`, compared as a set of `(relative path, code)` pairs with both roots resolved rather than by total.
- `node .github/scripts/assert-actions-pinned.mjs`: exit 0, 85 `uses:` references.

## The completeness gate stays opt-in, and the blocker is a real coverage gap

`fail_on_incomplete_scanners` remains **off by default**. It was flipped on during this work and reverted, because this repository cannot currently pass its own gate — and that is a measurement, not caution.

`cdk-nag` evaluates **6 of its 10 targets** here. #559 reports that as `PASSED (4 of 10 targets unevaluated)`, and with the gate on it becomes exit 1, so every `scan` leg fails. Measured on x86 Linux locally as well as on arm64 and Windows in CI, so it is not a platform artifact:

```
ash scan --scanners cdk-nag   →  Scan incomplete: cdk-nag (PASSED (4 of 10 targets unevaluated))
                                 ERROR (1) Exiting because the scan was incomplete
```

#559 documented the underlying fact first, in `incomplete_scanners`' own docstring: *"Measured on this repository under its own config: cdk-nag attempts 10 targets and fails 4 ... and the gate exited 0. **Two of those four are real CloudFormation templates that went unscanned.**"*

**That is the blocker on enabling the gate, and it is a live coverage gap rather than noise** — two CloudFormation templates going unscanned by a security scanner. It lives in `cdk_nag_wrapper.py`, outside this PR's files, so it is not fixed here. A gate whose first act is to fail every scan of its own repository teaches operators to pass `--no-fail-on-incomplete-scanners`, which is worse than shipping it opt-in.

### Nothing in this branch causes the incomplete scan, and that took measuring

Read quickly, this PR looks like gate machinery sitting next to a known-incomplete scan, and the obvious inference is that the branch caused the incompleteness. It did not. Three separate investigations here reached that wrong conclusion before the measurements came in, so both facts are stated with their counts:

- **cdk-nag's 4-of-10 target coverage is pre-existing on `main`.** Documented first by #559, in `incomplete_scanners`' own docstring, before this branch existed.
- **`AwsSolutions-EC26` and `AwsSolutions-EC27` fail to evaluate on every ref, and that is the fixture's purpose.** Measured with `-n 0` on the same machine, `tests/integration/scanners/test_cdk_nag_real_pack.py -k TestRuleThatCouldNotBeEvaluated`: `could not be evaluated` appears **10 times at `cc20a9938d`** and **10 times at this PR's head**, naming the same two rules, 5 passed both times. `unevaluatable_template` is built so `NagRules.resolveIfPrimitive` raises on a parameter `Ref`; one of that class's own assertions fails if the rules ever stop throwing — *"no rule reported as un-evaluated, so this fixture no longer reproduces the defect."*
- **No extra scan pass.** Instrumenting both entry points and counting per test gives an identical tally on both refs: 1 wrapper call, 5 `CdkNagScanner.scan` calls, with the one test that scans twice (scan, suppress, rescan) doing so on both.

An earlier 2.9s-versus-11s runtime comparison suggested otherwise and was an artifact: the same tests at the same ref measured `in 11.52s` then `in 3.27s`, a cold jsii/cdk package cache on first invocation.

### Two prerequisites for enabling the gate, not one

1. **cdk-nag's coverage gap**, above — `cdk_nag_wrapper.py`.
2. **`cfn-nag`, `grype` and `syft` report MISSING on Windows** — Ruby and two Anchore Go binaries, not provisioned there. Closing prerequisite 1 alone will not let a Windows scan pass. Each needs a SKIPPED-versus-MISSING decision: declared-unsupported-and-tolerated, or a real provisioning gap that should fail. (This one is from a colleague's reading of `scan (python-local, windows-latest)`; I did not read that log myself. It is consistent with the four-MISSING measurement quoted in `test_the_default_is_off_and_the_opt_in_is_what_gates`.)

   For `cfn-nag` specifically, part of that MISSING is a **bug rather than a decision**, and it is fixed in this PR. The `Install cfn-nag` step in `run-ash-security-scan.yml` ran `sudo BUNDLE_GEMFILE=... bundle install`, which fails with `sudo: bundle: command not found`: `sudo` replaces `PATH` with `secure_path` from `/etc/sudoers`, and Ruby on the GitHub ubuntu images is not under any directory that lists. So the step that exists to stop cfn-nag being MISSING had never worked. Its own comment predicted this shape — *"if it were wrong, the scan would not fail here, it would report cfn-nag MISSING"* — and that is what happened: this PR's own `ash` scan died four steps downstream, on `Path does not exist: .ash/ash_output/reports/ash.ghas.sarif`, with the scan never having run. Both `bundle` and `gem` are now resolved to absolute paths as the runner user before being handed to `sudo`, with a guard that names the failure if Ruby is absent entirely. So whoever revisits prerequisite 2 should re-measure cfn-nag's status **after** this fix rather than from the earlier logs.

Both pre-date this PR and were silently tolerated, which is the point: the gate is correct and the repository does not currently pass it.

What this PR does ship is the mechanism, working and tested through the explicit opt-in: `--fail-on-incomplete-scanners` gates on ERROR, MISSING, an unrecognized status, partial target coverage, and now a run where no scanner reached a verdict at all. #559's reporting of the gap is deliberately untouched — `ash.summary.md` still carries *"cdk-nag: evaluated 6 of 10 target(s); 4 could not be evaluated"* under **Incomplete coverage**. Visibility was the half worth keeping; gating is the half that cannot ship until the gap closes.

## Out of scope, found while verifying: a skip path that hard-fails instead of skipping

Not fixed here, and named because it is the same defect class this PR is about — an outcome that misreports what happened. It belongs to the owner of the file, not to this change.

`cdk-nag` is an optional extra (`[project.optional-dependencies] cdk`) with no `[tool.uv]` block setting default extras, so a plain `uv sync` does not install it. There is an intended skip-by-default path for that, and it is implemented once: `tests/integration/scanners/test_cdk_nag_real_pack.py:29-33` imports `cdk_nag` inside a `try`, skips on `ImportError`, and hard-fails instead when `ASH_REQUIRE_CDK_EXTRA` is set — which `.github/actions/run-unit-tests/action.yml:45` sets, so CI cannot silently stop testing that path.

That guard does not reach `tests/unit/utils/test_cdk_nag_unevaluated_rule.py:339`, which is a bare `import cdk_nag` with no `importorskip`, no `skipif`, and no reference to the variable anywhere in the file. So in a default local environment that test reports a **failure** where the design intends a skip. It is not the variable being set that causes it; that file simply has no fallback. Reproduced at a clean `cc20a9938d` with none of this PR's work present, so it is pre-existing on main and untouched here.

### Two things a reviewer will trip over in this PR's CI

**`grep FAILED` returns hits on a fully green log.** `FAILED` is a `ScannerStatus` member, so it appears in parametrization ids: **18 parametrizations across 8 test functions**, including `test_a_single_high_finding_across_the_ladder[HIGH-FAILED]`, `test_the_status_branch_table[flags8-FAILED]`, and `test_ran_statuses_are_exactly_the_two_that_mean_the_scanner_executed[FAILED]`, the last added here. Read the summary line, not the grep.

**A jsii package-cache race can redden one Windows Python version while the others pass.** Seen on `PyTest - windows-latest x86_64 Python 3.11`:

```
RuntimeError: EEXIST: file already exists, open
  'C:\Users\runneradmin\AppData\Local\AWS\jsii\package-cache\aws-cdk-lib\2.268.0\...lock'
```

Parallel test workers racing on a shared cache directory — infrastructure, not code, and not specific to any Python version. A red 3.11 leg beside green 3.10/3.12/3.13 is this, not a version-specific bug.


nikmeiser (approved):
Opened issue for residual issue - https://github.com/awslabs/automated-security-helper/issues/579

PR #580, opened 2026-09-17
fix(sarif): preserve rule severity during aggregation

### LINKED ISSUE #577: Summary severity counts mislabel HIGH findings as CRITICAL (_resolve_result_severity maps SARIF level=error → critical)
## Summary

The aggregated severity counts map SARIF `result.level == "error"` **unconditionally to `critical`**, so HIGH-severity findings (CVSS 7.0–8.9) are counted as CRITICAL. The same report's **Detailed Findings** section, `ash.flat.json`, and `ash.csv` correctly show those findings as **HIGH** — so `ash.summary.md` contradicts itself, and the summary table / `summary_stats` / MCP `get_scan_summary` over-report critical and under-report high.

**Affected version:** 3.7.1 (also present on `main` as of filing).

## Root cause

`automated_security_helper/utils/sarif_utils.py::_resolve_result_severity`:

```python
def _resolve_result_severity(result) -> str:
    if result.properties:
        ...
        issue_sev = props.get("issue_severity", "").upper() ...
        if issue_sev in ("CRITICAL", "HIGH", "MEDIUM", "LOW", "INFO"):
            return issue_sev.lower()
    if result.level:
        match str(result.level).lower():
            case "error":   return "critical"   # <-- error can be HIGH or CRITICAL
            case "warning": return "medium"
            case "note":    return "low"
    return "info"
```

Called from `core/scanner_statistics_calculator.py::extract_sarif_counts_for_scanner` to build the per-scanner `severity_counts`.

For grype findings:
1. The aggregated result's PropertyBag has **no `issue_severity`** (only `tags` / `scanner_name` / `scanner_details` / `scanner_version`), so the first branch is skipped.
2. It falls back to `result.level` and maps **`error → critical`**.
3. It never consults the rule's **`security-severity`** (CVSS numeric) that would distinguish high (7.0–8.9) from critical (≥9.0). That field is present in the native grype SARIF but is **dropped/`None` in the aggregated `ash.sarif`**.

Result: every `error`-level grype finding is bucketed as `critical`, regardless of actual CVSS.

## Reproduction

1. Scan a project with grype findings in the CVSS 7.0–8.9 range and none ≥9.0. Example npm advisories:
   - `brace-expansion` `GHSA-3jxr-9vmj-r5cp` (CVSS 7.7)
   - `vite` `GHSA-fx2h-pf6j-xcff` (CVSS 8.2)
   - `brace-expansion` `GHSA-rgw5-rvv9-x895` / `GHSA-mh99-v99m-4gvg` (CVSS 7.5)
2. Open `.ash/ash_output/reports/ash.summary.md`.

**Observed** — the Scanner Results table:

| Scanner | Critical | High | Medium | Actionable | Result |
| --- | ---: | ---: | ---: | ---: | --- |
| grype | **4** | **0** | 3 | 7 | FAILED |

…while the **Detailed Findings** section *in the same file* lists those 4 findings as `Severity: HIGH`.

Cross-checks (all agree the 4 are HIGH, not CRITICAL):
- `reports/ash.flat.json` → `HIGH`
- `reports/ash.csv` → `"A high vulnerability in npm package: ..."`
- `scanners/grype/source/results_sarif.sarif` → `security-severity` 7.5 / 8.2 / 7.5 / 7.7 (all in the HIGH band)
- `ash_aggregated_results.json` → `summary_stats.severity_counts` = `{critical: 4, high: 0, medium: 3}` (wrong — same source as the table)

## Expected

| Scanner | Critical | High | Medium | Actionable | Result |
| --- | ---: | ---: | ---: | ---: | --- |
| grype | 0 | 4 | 3 | 7 | FAILED |

## Impact

- Summary table (`ash.summary.md`, `ash.summary.txt`), `ash_aggregated_results.json` `summary_stats`/per-scanner `severity_counts`, and the MCP `get_scan_summary` all over-report CRITICAL and under-report HIGH.
- Affects any scanner whose findings reach the `result.level` fallback without an `issue_severity` in the aggregated PropertyBag (grype confirmed; likely others).
- Threshold gating via `calculate_actionable_count` is unaffected at MEDIUM (both critical and high are actionable), but any critical-count-based reporting/alerting is incorrect.

## Suggested fix

In the `level` fallback, don't assume `error == critical`. Options:
1. Consult `security-severity` (CVSS) when present: `>=9.0 → critical`, `7.0–8.9 → high`, `4.0–6.9 → medium`, `0.1–3.9 → low`. This also requires preserving `security-severity` through SARIF aggregation (it is currently dropped).
2. And/or write the normalized per-scanner severity into `issue_severity` in the aggregated PropertyBag so the existing first branch resolves it correctly.

danielgaskins:
## Summary

Grype stores its numeric severity on SARIF rule definitions, while its results may only carry the broad `error` level. During aggregation, those results fall back to `error` and are counted as CRITICAL even when the rule score is in the HIGH range.

This change copies a valid rule-level `security-severity` value into the result's canonical `issue_severity` property before aggregation. Existing result-level severities remain unchanged, and scanners without valid numeric metadata continue to use the existing fallback.

The tests cover CVSS boundaries, lookup by rule ID and index, existing severities, invalid metadata, missing rules, and the complete Grype aggregation path.

## Related Issues

Closes #577

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] CI/build change

## Checklist

- [x] Documentation updated (not applicable for this internal behavior fix)

## Testing

- `python -m pytest -q -o addopts='' tests/unit/utils/test_sarif_severity_normalization.py tests/unit/utils/test_severity_mapping_fix.py`
- 27 tests passed
- Ruff checks and formatting passed for the new test module

## AI Generation Disclosure

Full disclosure: I used Codex to inspect the severity data path, implement the change, draft the tests, and run validation. I reviewed the implementation and test results before submitting it.


nikmeiser:
**Nice fix for the Grype inflation — but flagging a fail-open regression on `npm-audit` before this merges.**

`normalize_sarif_result_severities` runs over every run in the aggregated SARIF, not just Grype's, and `npm-audit` also hits both trigger conditions:

1. Its results carry a `severity` property but **no `issue_severity`** (`npm_audit_scanner.py:249-258`), so the "skip if already set" guard doesn't fire.
2. Its rule `security_severity` **defaults to `0` when the advisory has no CVSS score** (`npm_audit_scanner.py:219` → `via.get("cvss", {}).get("score", 0)`; `:285` sets `0` for transitive vulns). npm advisories often ship no CVSS number.

`_severity_from_security_score(0)` returns `INFO`, so combined with npm-audit's severity→level map (`:172-179`):

| npm severity | level | before this PR | after this PR (no CVSS) |
|---|---|---|---|
| critical | error | **CRITICAL** | **INFO** |
| high | error | **CRITICAL** | **INFO** |
| moderate | warning | MEDIUM | **INFO** |

So a **critical npm vulnerability with no CVSS score is downgraded CRITICAL → INFO**, which drops it under any `--severity-threshold` above INFO. CI is green but doesn't cover this — there's no test with a score-0 / `error`-level finding.

The `0 → INFO` mapping is reasonable on its own (CVSS 0.0 = "None"); the underlying issue is npm-audit using `0` as a *sentinel* for "no score," which this PR activates. A couple of ways to resolve:

- **npm-audit**: omit `security_severity` when there's no CVSS score (instead of defaulting to `0`) — then the rule has no score, normalization skips it, and the `level` fallback is preserved.
- **this PR**: make the normalization non-downgrading — `_severity_from_security_score(0)` returns `None`, and/or never resolve a result *below* what the `level` fallback would give.

Doing both is safest, plus a regression test for a score-0 / `error`-level result. Happy to help with either. The Grype half of the change is correct and worth keeping.

nikmeiser:
I've pushed a fix to this branch (`d7e98a42`). Verified the fail-open end to end: an `error`-level npm-audit result whose rule carries `security_severity: 0` (npm's default when an advisory ships no CVSS number, including all transitive vulns) was normalized to `INFO` and dropped under the threshold.

Fix is defense-in-depth, and deliberately keeps the Grype half intact:

- **`_severity_from_security_score(0)` now returns `None`** instead of `INFO`, so a sentinel/zero score is skipped and the SARIF `level` fallback (`error → critical`) stands. Real positive scores are untouched, so Grype's intended `error → HIGH` refinement still works. (I went with your first option rather than a blanket non-downgrading guard, since Grype *legitimately* resolves below the level fallback — a non-downgrading guard would have broken exactly what this PR fixes.)
- **npm-audit now omits `security_severity` entirely** when there's no CVSS score, instead of emitting a misleading `0` (both the direct-vuln default and the hardcoded transitive `0`).
- **Regression test added**: a score-0, `error`-level result resolves to `critical`, and `0` joins the ignored-score cases.

`pytest` for the normalization + severity-mapping suites is green (28 passed incl. the new test). Grype half is unchanged.

nikmeiser:
Pushed follow-ups from a deeper review (`ca80163f`):

- **Per-scanner vs aggregate consistency (was a display split):** per-scanner counts were computed from un-normalized SARIF while the aggregate gate normalized, so a Grype finding could read CRITICAL in the per-scanner summary but HIGH in the total. Normalization now happens inside `get_severity_metrics_from_sarif`, so every counting path resolves the same severities as the gate (idempotent — the aggregate's explicit call is unaffected).
- **Non-canonical `issue_severity` no longer blocks the fix:** the normalizer skipped on any truthy `issue_severity`, but the resolver only honors the canonical bands — so a value like `"moderate"` blocked the rule-score correction and then silently fell to `level`. The skip is now gated on the canonical set (shared with the resolver).
- **Precedence policy documented as a deliberate, gate-affecting breaking change:** the rule-CVSS-over-`level` behavior stays **on by default** (intended), and it's now written up in a `### Breaking changes` CHANGELOG entry and a "How ASH determines a finding's severity" section in the scanner-statistics guide — noting it applies to every scanner (not just Grype), moves severity in both directions, and feeds the threshold gate and exit code.

Tests added for both fixes. Verified locally: 217 sarif/severity util tests pass, ruff clean.

awsmadi (approved):
lgtm

PR #586, opened 2026-09-17
fix(inspect): write suppressions to active config

### LINKED ISSUE #569: ash inspect findings writes suppression to root .ash.yaml instead of the config file used for the scan
### Summary
`ash inspect findings` (the suppression-authoring flow) writes new suppressions to a new `.ash.yaml` at the repository root, even when the scan was configured by a config file at a different path (e.g. `.ash/.ash.yaml`). The suppression lands in a file ASH did not scan with, so it has no effect and the finding keeps firing.

### Expected behavior
Suppressions created via `ash inspect findings` should be written to the SAME configuration file that was used for the scan (the resolved/active config path), not to a hardcoded root `.ash.yaml`.

### Actual behavior
A separate `.ash.yaml` is created at the repo root containing only the new `global_settings.suppressions` entry. The active config at `.ash/.ash.yaml` is left unchanged, so the suppression is silently ineffective and the file has to be hand-merged and deleted.

### Repro
1. Keep the ASH config at `.ash/.ash.yaml` (not repo root).
2. Run a scan; pick a finding.
3. Use `ash inspect findings` to add a suppression for it.
4. Observe a new `.ash.yaml` at the repo root with the suppression; `.ash/.ash.yaml` is untouched.

### Environment
- ASH version: 3.7.1

### Suggested fix
Resolve the active config path (same precedence ASH uses to LOAD config) and write the suppression back to that file; fall back to root `.ash.yaml` only when no config file was in use.

danielgaskins:
## Summary

- Discover the suppression config using the same precedence as the scan command.
- Add `--config` and `-c` support for nonstandard config locations.
- Pass the resolved path through the findings explorer to the suppression dialog.
- Report the actual file that received the suppression.
- Keep the root `.ash.yaml` fallback when no config exists.

## Related Issues

Closes #569

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] CI/build change

## Testing

- Findings explorer tests: 80 passed
- Config discovery and suppression tests: 58 passed
- CLI help verified
- Ruff formatting and static checks passed
- `git diff --check`

## Checklist

- [x] Documentation updated through CLI help


nikmeiser:
The config-discovery change here is a correct fix for #569 — no objection to the approach. But I'd hold merging this until a writer fix lands first, because **this PR amplifies a pre-existing bug it doesn't itself contain.**

`add_suppression_to_config` in `config/ash_config.py` (which this PR doesn't touch) full-rewrites the config via `yaml.safe_dump` whenever there's no `suppressions:` section yet — the common first-suppression case — which **strips every comment**. Today that's low-impact because inspect often writes to a fresh file. Once this PR points inspect at the *discovered active config* (in this repo, the heavily-commented `.ash/.ash.yaml`), the first `inspect` suppression would **delete that config's entire comment block**.

I've opened #593 to fix the writer (confine the comment-stripping rewrite to files with nothing to preserve; also fixes a latent YAML-escaping bug where a reason containing `\"` or `\\` produced invalid YAML). Recommend landing #593 first, then this merges safely on top.

One thing worth folding into this PR while it's open (it's in a file you already edit, `inspect_findings_app.py:109-116`, though not on lines your diff changes): a finding with no `file` becomes `path=self.finding.get(\"file\") or \"*\"` — a silent repo-wide suppression — and still attaches `line_start`/`line_end`, which is incoherent with `path=\"*\"`. Surfacing the resolved scope in the dialog and not attaching a line to `*` would close that gap.

PR #590, opened 2026-09-17
fix(ci): remove the ghcr.io pull and piped tool installers from the build path

awsmadi:
Two failures on run 35246976698, the merge-queue run for #587. Both are external dependencies the build did not need to have. The third change is the retry budget they were both measured against.

## A skipped step's image pull failed seven legs that never run it

`Code Coverage Report` in `.github/actions/run-unit-tests` is a Docker-based action whose `if` selects exactly one of the eighteen unit-test legs (Linux, py3.12, x86_64). A Docker action's image is pulled during the job's prepare phase, before any step's `if` is evaluated, so the condition did not stop the other legs from pulling it.

Measured across that run. Every Linux leg logged `Pull ghcr.io/irongut/codecoveragesummary:v1.3.0`, including `ubuntu-24.04-arm/py3.11` and `ubuntu-latest/py3.13` where the condition is false. macOS and Windows legs logged no pull at all, so the exposure is the eight Linux legs, one of which uses the step.

On `ubuntu-24.04-arm/py3.11` that pull failed DNS resolution three times, and because the prepare phase precedes execution, `Install dependencies`, `Run PyTest` and `Run cdk-nag integration tests` were skipped with `duration_ms=0`. The leg reported failure having run zero tests.

coverage.py renders Markdown itself and `pytest-cov>=7.1` already pins it, so the summary needs no container. `--fail-under=0` is load-bearing: `.coveragerc` sets `fail_under=80`, `coverage report` exits 2 below it, and this step must not become a second gate.

## Three tools were installed by piping a vendor script into a shell

`utils/tool_downloads.py` has carried pinned SHA256 digests for syft, grype and trivy since #568, and its docstring names `curl … | sh` as the alternative it rejected while naming this image as the place still doing it.

It failed on availability too. anchore's installer could not resolve `get.anchore.io` (`HTTP status=000`), fell back to `github.com`, got the 302 that every release download answers with, and treated the redirect as a failure — so the tarball was never written, its checksum "did not verify" against a file that did not exist, and `RUN syft --version` failed with exit 127. The guard at that line is correct and stays.

`assets/install-pinned-tool.py` resolves the asset from the pinned table, follows the redirect, verifies the digest, and refuses to install on a mismatch with a distinct exit code so `with-retry` does not retry an integrity failure.

It resolves the pins rather than calling `download_utils.install_pinned_tool`, which is the same operation: importing that pulls in 19 ASH modules including the pydantic-backed config, so it cannot run before the wheel is installed — and the wheel lands after these three tools. Moving it earlier would put the tool downloads downstream of every source change and re-fetch all three on every build, increasing exposure to the failure being fixed.

The shipped Dockerfile is generated by `hatch_build.py` and its build context is the assets directory, so `hatch_build` stages the two files it now needs and rewrites their COPY paths. Without that, container mode on an installed ASH would fail on a COPY outside its context.

## The retry budget was shorter than the outage

Both failures were runner egress, and `with-retry.sh`'s three attempts spend about 15 seconds. The syft attempts ran at 16:41:55, 16:42:01 and 16:42:12 and all three failed to resolve a host; the ghcr.io pull on the other runner failed for 2m16s straight. Five attempts sleep 5s, 10s, 20s and 40s. `download_utils`' default moves with it, since the two are documented as mirroring each other.

## Verification

- Full suite: 7661 passed, 0 failed, coverage 93.45% against the 80% gate.
- Both container variants build (root `Dockerfile` and the generated `assets/Dockerfile` in its own context), with all three tools at the pinned versions and each digest logged.
- A tampered digest exits 3 and leaves the destination empty.
- Artifact gate passes on wheel and sdist; its self-test still fires all 18 detectors.
- Lint job's schema regeneration and docs-freshness checks pass.

Neither pinned version changed: `TOOL_VERSIONS` already matched the Dockerfile's `ARG` lines for all three tools on both amd64 and arm64, and a new test asserts they stay equal.

## Not addressed

`.github/workflows/run-ash-security-scan.yml` installs grype and syft with `curl … | sudo sh`, and the trivy plugin docs document the vendor's piped installer. Same pattern, different blast radius, and neither is on the path that failed.


rafaelpereyra (approved):

awsmadi:
Folded into #587 and closing this in favor of it.

All five commits were cherry-picked onto `ci/main-push-gate`, which is now at `a562f463`:

- `dc5afb5b` remove the ghcr.io pull and piped tool installers from the build path
- `f29655d1` stop scanning the build-generated copies of the pinned tool table
- `2a226a05` retry and pin the last three unretried network installs
- `4859ac52` stop the Dockerfile assuming a shell podman does not give it
- `15f58ea3` list the new pinned-tool test in the per-file B404 allowlist

Nothing was dropped and nothing was rewritten: `68075150` remains an ancestor of the new head, the picks applied without conflict, and the full unit suite is green at `a562f463` with 7701 passed.

The reason for combining them is that these fixes are what stop #587 from being ejected from the merge queue. That pull request was ejected twice by the two failures fixed here, and while separate they had to land in a fixed order, with #587 waiting on this one. #587's description now covers all three concerns separately so the wider diff stays reviewable.

PR #594, opened 2026-09-18
test(cdk-nag): locate cdk-nag without importing it, killing a Windows jsii flake

nikmeiser:
## The flake

`tests/unit/utils/test_cdk_nag_unevaluated_rule.py::TestTheMarkerStillMatchesTheInstalledCdkNag::test_the_prefix_appears_verbatim_in_the_installed_distribution` has intermittently failed on **Windows** across several PRs (#562, #570, #591, #592, …) with:

```
RuntimeError: EEXIST: file already exists, open
'C:\Users\runneradmin\AppData\Local\AWS\jsii\package-cache\aws-cdk-lib\<ver>\<sha>.lock'
```

## Root cause

The test does `import cdk_nag` **only to find the package directory** (`Path(cdk_nag.__file__).parent`), then reads static `*.tgz` files from it. But importing `cdk_nag` runs its `__init__`, which imports `aws_cdk._jsii` and starts a **jsii kernel** that extracts `aws-cdk-lib` into a shared per-user cache (`%LOCALAPPDATA%\AWS\jsii\package-cache`). Under `pytest-xdist` on Windows, two workers hit that first extraction simultaneously and race on the cache's `.lock` file — `jsii`'s `lockSyncWithWait` opens it exclusively, so the loser gets `EEXIST`. It's a concurrency bug in jsii's Windows cache extraction; the test doesn't need the runtime at all.

## Fix

Locate the distribution with `importlib.util.find_spec("cdk_nag")`, which resolves the module **without executing it** — no `aws_cdk` import, no jsii kernel, no cache extraction, no race. The test still reads the same on-disk tarballs and makes the same assertions.

The other import in the test (`_UNEVALUATED_DESCRIPTION_PREFIX` from `cdk_nag_wrapper`) is jsii-free — that module only imports `cdk_nag` inside functions, not at module load.

## Verification

In a venv with `cdk-nag` + `aws-cdk-lib` installed, the locate-and-read logic:
- finds the tarball and `lib/nag-pack.js` (with `addViolation`) — assertions still resolve;
- leaves **`aws_cdk` and `cdk_nag` out of `sys.modules`** — proving the jsii kernel never starts, so the Windows race cannot occur.

No behavior change to what the test asserts; it still fails loudly if cdk-nag rewords the marker or if cdk-nag isn't installed.


awsmadi (approved):
lgtm

PR #622, opened 2026-09-20
fix(inspect): write suppressions to active config

### LINKED ISSUE #569: ash inspect findings writes suppression to root .ash.yaml instead of the config file used for the scan

danielgaskins:
## Summary

Fixes suppression authoring when the ASH config lives in `.ash/` or at a custom path.

- Uses ASH's existing config discovery order to select the suppression target
- Adds `--config/-c` and `ASH_CONFIG` support for nonstandard config locations
- Passes the selected path into the suppression dialog
- Names the updated file in the success message

## Related Issues

Closes #569

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] CI/build change

## Checklist

- [x] Documentation updated (the CLI help documents the new option)

## Testing

- 80 inspect command tests passed
- 10 configuration discovery tests passed
- Verified that the command help exposes `--config/-c` and `ASH_CONFIG`

The full unit suite couldn't collect locally because the current environment is missing optional dependencies including `mcp`, `starlette`, `boto3`, and `botocore`.


rafaelpereyra (approved):
LGTM, thanks for the contribution @danielgaskins !!

